PR #4, opened 2023-07-04
Feature speed up epc count

JoshAdamPowell:
use raw sql to speed up querying the amount of epc rating rows.


CheshireSwift (approved):
Just out of curiosity, don't suppose there's a test for this page? (By design, it'd still pass with this refactor, but it'd be good to have that confirmation 😄)

PR #20, opened 2023-07-05
PC-346: Add dynamic backlink to accessibility statement page

chidin194:


help_to_heat/templates/frontdoor/accessibility-statement.html
```diff
@@ -5,8 +5,13 @@
 {% endblock %}
 
 {% block content %}
-<div class="govuk-grid-row">
-  <div class="govuk-grid-column-two-thirds">
+
+    {% if url %}
```
CheshireSwift, line 9: Is this `if url` necessary/doing anything useful? 🤔

CheshireSwift (approved):
One query, otherwise lgtm (and not too worried about the query assuming this has tested okay locally).

chidin194:
Changed this and amended the indents/formatting, although I'm not sure how much of a concern this is?

PR #25, opened 2023-07-07
Debug

JoshAdamPowell:
I think the autoformatter has messed up some things. The important changes are: in docker-compose and Dockerfile we're exposing 5678. added launch.json, and the couple of lines to settings.py


CheshireSwift (approved):
Slightly bemused by the requirements rearrangements, but fine with your changes as stated.

PR #26, opened 2023-07-10
PC-373: Add session id to response headers

JoshAdamPowell:
Adds session ID to responses. (Fun fact the session ID is generated on the homepage even though you can't see it yet)


CheshireSwift (approved):
I'd still rather we'd use a header that isn't Cloud Foundry specific and is more intuitively named (e.g. `x-request-id`), but in the interests of getting the load testing planned/done, this is fine.

PR #28, opened 2023-07-11
PC-393: Add role and tab index to NI gov link

chidin194:
Similar to the PR for PC-392, I've added a role and tabindex of 0, although the 'link' was already being read out by the screen reader. 

I wonder if I'm testing this correctly and if I should liaise with the reporter to reproduce exact steps? There's not much detail on the ticket.

### DIFF
--- a/help_to_heat/templates/frontdoor/northern-ireland.html
+++ b/help_to_heat/templates/frontdoor/northern-ireland.html
@@ -9,7 +9,7 @@
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
       <h1 class="govuk-heading-l">This service is not available for homes in Northern Ireland</h1>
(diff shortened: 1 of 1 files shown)


CheshireSwift:
>  'link' was already being read out by the screen reader.
Yeah, I'm curious about this change. It's not obvious to me why the screen reader wouldn't be reading this link, and I really don't want to start messing with taborder on a hunch - that tends to make accessibility worse not better.

> liaise with the reporter to reproduce exact steps
Yes please.

CheshireSwift (commented):
>  'link' was already being read out by the screen reader.

Yeah, I'm curious about this change. It's not obvious to me why the screen reader wouldn't be reading this link, and I really don't want to start messing with taborder on a hunch - that tends to make accessibility worse not better.

> liaise with the reporter to reproduce exact steps

Yes please.

chidin194:
@CheshireSwift confirmed by Adrian that this is the fault of the screen reader simulator in use, and so I'm closing this PR

PR #29, opened 2023-07-11
Pc 344 invalid email 500 bug

JoshAdamPowell:
@CheshireSwift Do you know why it's picking up which files are changed but just taking the whole thing? it's very annoying.


CheshireSwift:
My guess would be line endings in some way?

help_to_heat/portal/email_handler.py
```diff
@@ -74,24 +77,33 @@ def _send_token_email(user, subject, template_name, from_address, url_name, toke
...
+        response = send_mail(
+            subject=subject,
+            message=body,
+            from_email=from_address,
+            recipient_list=[user.email],
+        )
+        return response
+    except Exception as err:
+        logger.error("An error occured while attempting to send an email.")
+        logger.error(err)
```
CheshireSwift, line 90: Presumably this is just swallowing the error. Might we want to surface the error to whatever called this to handler, or are we confident that logging and throwing the attempt away is correct?

help_to_heat/portal/email_handler.py
```diff
@@ -74,24 +77,33 @@ def _send_token_email(user, subject, template_name, from_address, url_name, toke
...
-    return response
+    try:
+        response = send_mail(
+            subject=subject,
+            message=body,
+            from_email=from_address,
+            recipient_list=[to_address],
+        )
+        return response
+    except Exception as err:
```
CheshireSwift, line 103: (Same question as above.)

CheshireSwift (commented):
Query about intended behaviour, but assuming we're happy swallowing the error then I'm happy to approve.

PR #35, opened 2023-07-12
Pc 277 add subtext to walls question

chidin194:
image


CheshireSwift:
Currently pointing at wrong branch 🙂

CheshireSwift (changes requested):
Yeah, I suspect you can simplify the rest of the list back down to just being strings, with only the hint-requiring one being a full dictionary.

chidin194:
> Yeah, I suspect you can simplify the rest of the list back down to just being strings, with only the hint-requiring one being a full dictionary.

Changes made :)

PR #37, opened 2023-07-12
Pc 394 add back link to privacy policy page

chidin194:
Made changes equivalent to the accessibility statement page, and tested that the same behaviour can be seen


CheshireSwift (approved):
lgtm 👍

PR #40, opened 2023-07-13
PC-359: dont cache sensitive data

JoshAdamPowell:
I have no idea if this is particularly "django-ic" but it seems fairly sensible to me

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -156,7 +156,7 @@ class PageView(utils.MethodDispatcher):
         }
         response = render(request, template_name=f"frontdoor/{page_name}.html", context=context)
         response["x-vcap-request-id"] = session_id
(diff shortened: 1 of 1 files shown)


help_to_heat/frontdoor/views.py
```diff
@@ -156,6 +156,9 @@ def get(self, request, session_id, page_name, data=None, errors=None, is_change_
         }
         response = render(request, template_name=f"frontdoor/{page_name}.html", context=context)
         response["x-vcap-request-id"] = session_id
+        if "sensitive" in context and context["sensitive"] == True:
```
CheshireSwift, line 159: `if foo == True` =>`if foo`? 🙂

CheshireSwift (commented):
I am likewise not fluent in Django, but this looks like a reasonable option to me.

CheshireSwift (approved):

PR #42, opened 2023-07-13
Pc 366 holding page

chidin194:
I think we might want to do something slightly different for the link in the page heading banner for 'Check eligibility for Great British Insulation Scheme. Currently, in base.html it redirects to the homepage but I assume if this page is being shown then the homepage won't be available. We could either change the href to just redirect back to this holding page, or only render it if the page isn't the holding page? It doesn't appear in the Figma designs [here](https://www.figma.com/file/hJDD86FSndWDG9zo8RClmm/C-team-project?type=design&mode=design). 

Alternatively, if the plan is to make '/' redirect to the holding page when it's in use, please ignore!

Not sure if the same logic could be applied to the feedback section in that a user wouldn't be able to review the service to offer feedback, but that is included in the designs. 

image


LadunOmideyiSoftwire (changes requested):
They have un-confirmed the text so we'll need to confirm and revise before pushing.

LadunOmideyiSoftwire:
Let's not merge til end of day. We need this to go to prod but not to staging or dev.

PR #58, opened 2023-07-20
Waitress threads 16 -> 64

LukaszDygon:
I wasn't sure whether this is the right branch, but I followed the flow in https://softwiretech.atlassian.net/wiki/spaces/Support/pages/20519190529/GBIS+Help+to+Heat+Development+process

My only change is waitress threads 16 -> 64

The reason behind it is waitress queuing during load testing.
This increase may lead to general slow down, as there are more threads than thread vcores on our compute insteance, but we will find out how much impact this has during load testing


LukaszDygon:
Will do

CheshireSwift:
Oh, for load testing we'll also want to cherry-pick these changes onto prod (`main` branch). That will probably need a new PR (I believe it's protected).

PR #61, opened 2023-07-20
increase waitress threads 16 -> 64

LukaszDygon:
single hotfix commit. Already on staging

### DIFF
--- a/docker/web/start.sh
+++ b/docker/web/start.sh
@@ -13,4 +13,4 @@ echo "Suppliers check completed"
 
 echo "Starting app"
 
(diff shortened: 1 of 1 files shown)


JoshAdamPowell:
I have this email from Ed:
The app is mostly likely io bound not cpu bound (as there's nothing intensive to compute), which likely means slow database (as thats the main io), unless you are trying to serve the static files from waitress, in which case that's probably the cause of the io.  (Obviously static files should be served by a web server (or CDN) not an application server).

Increasing the number of threads is unlikely to help.  There's already 4x the number of cores, so adding more won't have any effect (and may even slow it down), but again, experiment to find out.


I'm happy to approve this though, I've got no idea :P

JoshAdamPowell (approved):

PR #70, opened 2023-07-21
Scottish epc load command

LukaszDygon:
For now this is WIP, worth adding link to this to Swiki if approved, because imo it automates and simplifies at least the creation of data to be loaded into the DB

### DIFF
--- a/utils/README.md
+++ b/utils/README.md
@@ -1,0 +1,27 @@
+# Utility scripts
+
+## Install
+
+```sh
+pip install virtualenv`
+python -m virtualenv .venv
+.\.venv\Scripts\activate
+pip install -r requirements.txt
+```
+
+## Scripts
+
+### update_scottish_epc
+
+Currently WIP
+
+For now this creates a CSV file containing scottish EPC data ready to be loaded into the DB manually,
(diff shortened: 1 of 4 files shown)


utils/README.md
```diff
@@ -0,0 +1,27 @@
...
+### update_scottish_epc
+
+Currently WIP
+
+For now this creates a CSV file containing scottish EPC data ready to be loaded into the DB manually,
+as described in [Swiki article](https://softwiretech.atlassian.net/wiki/spaces/Support/pages/20520173599/Common+Tasks#How-to-actually-do-an-EPC-data-dump-for-fun-%26-profit).
+
+TODO: Add automatic loading to DB.
+
+The data is deduplicated (using latest only), sorted in order of files, with records without EPC number dropped.
```
JoshAdamPowell, line 23: it's records without uprns that are dropped, UPRNs are property IDs (Unique Property Reference Number)

utils/README.md
JoshAdamPowell, line 1: The process is near identical for English EPCs so we can probably modify it a little to work on either type :)

CheshireSwift:
Will check more thoroughly next week, but at a glance this looks real nice, ty 🙏

CheshireSwift:
Closing old PR (but retaining in GitHub for reference if required).

PR #79, opened 2023-07-24
PC-NONE: Use pytest instead of nose for test running

JoshAdamPowell:


docker/tests/Dockerfile
```diff
@@ -25,6 +25,8 @@ RUN chmod +x /start.sh
 COPY ./docker/tests/check-code.sh /check-code.sh
 RUN chmod +x /check-code.sh
 
+RUN python3 -m pip install -U pip pytest
```
CheshireSwift, line 28: Following the pattern from `nose`, I think it fits better with existing infra to have this listed in the dev requirements, and installed from there rather than a separate command in the dockerfile.

CheshireSwift (changes requested):
Small change to how we import pytest, otherwise lgtm

CheshireSwift (approved):

PR #87, opened 2023-07-27
Pc 413 bulb warning page

xnw9:
https://beisdigital.atlassian.net/browse/PC-413

Supplier page - Bulb -> warning page -> contact details page
                          Bulb <-                        <- contact details page
                       - others -> contact details page
                         others <- contact details page

`Bulb, now part of Octopus Energy` will always be shown as the supplier except
- on success page
- in referral email
- in the actual referral

image
image
image

Tests have been fixed.

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -139,21 +139,21 @@ def get_prev_next_urls(session_id, page_name):
 
 def get_supplier_and_add_comma_after_bulb(session_id):
     supplier = interface.api.session.get_answer(session_id, "supplier")["supplier"]
-    if supplier == "Bulb, now a part of Octopus Energy":
+    if supplier == "Bulb, now part of Octopus Energy":
         supplier = supplier + ", "
     return supplier
 
 
 def get_supplier_and_change_bulb_to_octopus(session_id):
     supplier = interface.api.session.get_answer(session_id, "supplier")["supplier"]
-    if supplier == "Bulb, now a part of Octopus Energy":
+    if supplier == "Bulb, now part of Octopus Energy":
         supplier = "Octopus"
     return supplier
 
 
 def get_session_data_and_change_bulb_to_octopus(session_id):
     session_data = interface.api.session.get_session(session_id)
-    if session_data["supplier"] == "Bulb, now a part of Octopus Energy":
+    if session_data["supplier"] == "Bulb, now part of Octopus Energy":
         session_data["supplier"] = "Octopus"
     return session_data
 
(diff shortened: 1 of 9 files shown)


help_to_heat/frontdoor/views.py
```diff
@@ -137,6 +137,27 @@ def get_prev_next_urls(session_id, page_name):
     return prev_page_url, next_page_url
 
 
+def get_supplier_and_add_comma_after_bulb(session_id):
```
CheshireSwift, line 140: Inclined to say we should wrap these three methods up in a class (probably one that takes `session_id` in the constructor). Maybe `BulbSupplierConverter`?

No real functional change, but just grouping them together.

Might even be worth moving that class to `interface`, including the current `sanitise_data_for_referral` logic as `replace_bulb_with_octopus_in_data` or something, and then `sanitise` just becomes `data = replace_bulb_with_octopus_in_data(data); return data`. Just since it's hanging on the same logic as the other three.

None of this is essential, what's here is fine, but would be good to keep related stuff grouped more explicitly 🙂

tests/test_frontdoor.py
```diff
@@ -165,7 +165,7 @@ def test_happy_flow():
     referral.delete()
 
 
-def _do_happy_flow(supplier="Foxglove"):
+def _do_happy_flow(supplier="EON"):
```
CheshireSwift, line 168: Guessing this change fixes some unhappy tests?

tests/test_frontdoor.py
```diff
@@ -190,7 +190,13 @@ def _do_happy_flow(supplier="Foxglove"):
     page = form.submit().follow()
 
     assert page.has_one("h1:contains('Select your home energy supplier from the list below')")
-    page = _check_page(page, "supplier", "supplier", "EON")
+
+    page = _check_page(page, "supplier", "supplier", supplier)
+
+    if supplier == "Bulb, now a part of Octopus Energy":
```
CheshireSwift: Nice little branch handling 👍

tests/test_frontdoor.py
```diff
@@ -893,16 +892,14 @@ def test_address_validation():
...
+
     models.Supplier(name="Octopus").save()
 
     session_id = _do_happy_flow(supplier=supplier)
 
-    data = interface.api.session.get_session(session_id)
-
-    assert data["supplier"] == "Octopus", data["supplier"]
+    referral_email_text = utils.get_latest_email_text("freddy.flibble@example.com")
+    assert "Your details have been submitted to Octopus." in referral_email_text
```
CheshireSwift, line 902: Good test.

CheshireSwift (changes requested):
Looks good. I think I'd like to extract some of the repetitive and conceptually related logic to a class just to keep it grouped.

tests/test_frontdoor.py
xnw9: Yeah - Foxglove is commented out in supplier options :)

help_to_heat/frontdoor/views.py
xnw9: Cool, let me give it a try :)

CheshireSwift (changes requested):
Couple more bits of minor tidying, otherwise lgtm.

xnw9:
I also tried 
image
:thinkies:

CheshireSwift:
Quite like the fully edited version screenshotted 🙂

CheshireSwift (approved):
🎉

PR #90, opened 2023-07-28
Pc none hide secrets ssssshhhhhh

JoshAdamPowell:
envs/web is already in the .gitignore, but I've renamed it now


CheshireSwift (approved):
I trust the env.template works fine in the non-local environments where it's getting used (portal checks, tests)?

If so, lgtm

PR #91, opened 2023-07-31
Pc 428 mock api

xnw9:
https://beisdigital.atlassian.net/browse/PC-428

- Used a mock api in real code
- Combined `get_by_postcode` and `get_by uprn`
- Refactored the mock api in test utils


help_to_heat/frontdoor/interface.py
```diff
@@ -221,9 +219,8 @@ def find_addresses(self, building_name_or_number, postcode):
 
     @with_schema(load=GetAddressSchema, dump=FullAddressSchema)
     def get_address(self, uprn):
-        api = osdatahub.PlacesAPI(settings.OS_API_KEY)
-        api_results = api.uprn(int(uprn), dataset="LPI")["features"]
-        address = api_results[0]["properties"]["ADDRESS"]
+        api_results = MockOSApi(settings.OS_API_KEY).get_by_uprn(int(uprn), "LPI")["results"]
```
xnw9, line 222: Before, it was using `OSApi` in `.os_api` to get postcode results, but using `osdatahub.PlacesAPI(settings.OS_API_KEY)` to get uprn results. I changed it to use the same api.

tests/test_eligibility.py
```diff
@@ -288,7 +288,7 @@ def _check_page(page, page_name, key, answer):
     return _check_page
 
 
-@unittest.mock.patch("help_to_heat.frontdoor.interface.OSApi", utils.StubAPI)
+# @unittest.mock.patch("help_to_heat.frontdoor.interface.OSApi", utils.MockApi)
```
xnw9, line 291: I commented it out as we are using mock api in real code already. I changed `StubAPI` to the new `MockApi` so that we can just uncomment this line when we switch back to real api.

PR #95, opened 2023-08-01
PC-441: Persist db connection

JoshAdamPowell:


JoshAdamPowell:
I've only tested this in so much as it works locally. I think I'll have to work with the QA people to test if this actually helps :)

CheshireSwift (approved):
Obligatory check: presumably our PG instance is happy to have a connection open for 5 minutes?

Assuming that's okay, lgtm.

JoshAdamPowell:
Locally it's happy, I'll check on dev once we've merged

PR #103, opened 2023-08-03
Pc 380 cookie banner

xnw9:
https://beisdigital.atlassian.net/browse/PC-380

Added a reject button

image

Also fixed the warning after cookie banner disappeared
image


xnw9:
We currently don't have tests for cookie banner. Do we want any?

help_to_heat/templates/frontdoor/base.html
```diff
@@ -29,7 +29,7 @@
 </head>
 
 <body class="govuk-template__body" id="i-dot-ai">
-  {% if request.COOKIES['cookiesAccepted'] != "True"%}
+  {% if request.COOKIES['cookiesAccepted'] != "True" and request.COOKIES['cookiesAccepted'] != "False"%}
```
xnw9, line 32: We currently don't have tests for cookie banner. Do we want any?

static/main.js
```diff
@@ -17,14 +18,27 @@ window.onload = () => {
     cookieBanner.classList.add("visible");
   }
 
-  hideButton.addEventListener("click", () => {
-    const date = new Date();
-    date.setFullYear(date.getFullYear() + 1);
+  if (cookieBanner) {
```
JoshAdamPowell, line 21: what happens if a user has js disabled? IIRC gov websites have to work fine without it.

xnw9: Oh. None of the buttons will work 😅 I guess we need to refactor the whole thing..?!

LadunOmideyiSoftwire: @CheshireSwift ?

JoshAdamPowell: yeah I think so. It's probably fine is no js -> just disallow cookies and dont show the thing

xnw9: From Sam: The buttons need to be a form, probably one that does a POST to the server. There can be a hidden input on the form saying what the current page is that we can use to redirect back to the same place.

help_to_heat/templates/frontdoor/base.html
CheshireSwift: Yeah, we should get some in. The non-JS version should be relatively easy to test.

Ideally wanna check all of:
- if you accept, you get acceptance cookie saying yes
- if you reject, you get acceptance cookie saying no
- if you have acceptance cookie yes then you get other cookies
- if you have acceptance cookie no then you don't get other cookies

static/main.js
CheshireSwift: Yeah, no need for any JS here that I can see. HTML form, server handling.

CheshireSwift (changes requested):

help_to_heat/templates/frontdoor/base.html
xnw9: Update: we're not adding any new tests for now

help_to_heat/context_processors.py
```diff
@@ -1,8 +1,14 @@
...
 def add_settings_to_context(request):
     return {
         "DEBUG": settings.DEBUG,
         "space_name": settings.VCAP_APPLICATION.get("space_name", "unknown"),
     }
+
+
+def suppress_cookie_banner(request):
+    suppress_cookie_banner_value = os.environ.get('SUPPRESS_COOKIE_BANNER', False)
+    return {'SUPPRESS_COOKIE_BANNER': suppress_cookie_banner_value}
```
CheshireSwift, line 14: This should be part of the `add_settings_to_context` above, going via the settings file (same as DEBUG).

help_to_heat/templates/frontdoor/base.html
```diff
@@ -29,33 +29,37 @@
...
-          <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies we are using on our service</h2>
-          <div class="govuk-cookie-banner__content">
-            <p class="govuk-body">We will use necessary cookies when you use this service to help you move around the
-              site.</p>
-            <p class="govuk-body">We will also use analytic cookies using Google Analytics to tell us how you got to our
-              service, the pages you visit, and
-              how long you spent on them. Analytics cookies help us to improve our service. We do not allow Google to
-              use or share the
-              data about how you use this site and the data is stored anonymously.</p>
+  {% if not SUPPRESS_COOKIE_BANNER %}
```
CheshireSwift, line 32: I believe this can just be merged into the conditional below, to avoid needing two `if`s?

static/main.js
```diff
@@ -1,30 +0,0 @@
-document.body.className = document.body.className
-  ? document.body.className + " js-enabled"
-  : "js-enabled";
-
-window.GOVUKFrontend.initAll();
-
-window.onload = () => {
```
CheshireSwift, line 7: We still need this script, it's just this `onload` that should go away.

CheshireSwift (changes requested):
Mostly looks fine, couple of things we need to rework/fix 🙂

help_to_heat/templates/frontdoor/base.html
xnw9: Ah yes! It complains about formatting but still works 😂
image

static/main.js
xnw9: 😖Ahhh sorry!

help_to_heat/templates/frontdoor/base.html
```diff
@@ -29,33 +29,35 @@
...
-          <h2 class="govuk-cookie-banner__heading govuk-heading-m">Cookies we are using on our service</h2>
-          <div class="govuk-cookie-banner__content">
-            <p class="govuk-body">We will use necessary cookies when you use this service to help you move around the
-              site.</p>
-            <p class="govuk-body">We will also use analytic cookies using Google Analytics to tell us how you got to our
-              service, the pages you visit, and
-              how long you spent on them. Analytics cookies help us to improve our service. We do not allow Google to
-              use or share the
-              data about how you use this site and the data is stored anonymously.</p>
+  {% if not SUPPRESS_COOKIE_BANNER and (request.COOKIES['cookiesAccepted'] != "True" and request.COOKIES['cookiesAccepted'] != "False") %}
```
JoshAdamPowell, line 32: Is this not `request.COOKIES['cookiesAccepted'] is not None`?

JoshAdamPowell (approved):
one small thing

help_to_heat/templates/frontdoor/base.html
xnw9: It will complain about `None` or `Null` :(

JoshAdamPowell: ok :)

PR #104, opened 2023-08-04
PC-441: Add debug toolbar

JoshAdamPowell:


JoshAdamPowell:
I'm slightly concerned this will add the toolbar to all environments.

JoshAdamPowell:
ok I'm more confident now

CheshireSwift (approved):
lgtm. Obvs would be good to double check after our next staging deploy that we've not acquired a rogue toolbar 🙂

PR #115, opened 2023-08-10
Pc 447 move supplier page

xnw9:
https://beisdigital.atlassian.net/browse/PC-447

- Fix a bug along the way - before, if we go to supplier from summary page and select continue, we'll have to go through the whole journey again

### DIFF
--- a/help_to_heat/frontdoor/interface.py
+++ b/help_to_heat/frontdoor/interface.py
@@ -134,12 +134,12 @@ class BulbSupplierConverter:
     def get_supplier_and_replace_bulb_with_octopus(self):
         supplier = self._get_supplier()
         if self._is_bulb():
-            return "Octopus"
+            return "Octopus Energy"
         return supplier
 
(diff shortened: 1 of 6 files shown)


CheshireSwift (commented):
What's here looks okay; would be nice if the tests were easier to work with but that's probably a wider conversation, not to do with this PR.

It looks like `test_bulb_to_octopus` is failing in the pipeline but not when I pull `develop` locally; could that be related to these changes?

xnw9:
> What's here looks okay; would be nice if the tests were easier to work with but that's probably a wider conversation, not to do with this PR.
> 
> It looks like `test_bulb_to_octopus` is failing in the pipeline but not when I pull `develop` locally; could that be related to these changes?

Ahhh, it was passing before I pushed,, probably something was lost during rebasing, sorry :(

LadunOmideyiSoftwire (changes requested):
Can I ask about "Octopus Energy"? It should still be Octopus, no? Except for the sentence "Bulb, now part of Octopus Energy"

PR #118, opened 2023-08-14
PC-427: Remove epc not found page

AyaPK:
# PC-427: Remove sad flow when an EPC is not found
- Updates the EPC controller to render the benefits page if an EPC is not found
- Flow remains identical when an EPC is found
- Removes references to the EPC page in tests that don't expect to see it


JoshAdamPowell:
what is up with that commit history? Can you squash it please

help_to_heat/frontdoor/urls.py
```diff
@@ -3,6 +3,7 @@
 from . import views
 
 frontdoor_patterns = [
+    path("thankyou", views.holding_page_view, name="holding-page"),
```
JoshAdamPowell, line 6: is this duplicating line 8?

JoshAdamPowell (approved):

help_to_heat/frontdoor/urls.py
AyaPK: Ah, yes! (Seems to just be some weirdness from rebasing :D)

PR #120, opened 2023-08-14
PC-455: Update feedback thanks page

AyaPK:
# PC-455: Update feedback thanks page
- Updates the feedback page text
- Adds a button for a user to return to their applications
-----
image


JoshAdamPowell:
I think it needs to be a 'panel': https://design-system.service.gov.uk/components/panel/

PR #121, opened 2023-08-14
PC-401-csv-fix: Add formatting

xnw9:
https://beisdigital.atlassian.net/browse/PC-401

(Also make `accept_suggested_epc` empty as a part of https://beisdigital.atlassian.net/browse/PC-400 which is a ticket in Done - if this is not want we want I'll change it back😊)


help_to_heat/portal/download_views.py
```diff
@@ -77,12 +77,20 @@ def add_extra_row_data(referral):
     eligibility = calculate_eligibility(row)
```
xnw9, line 77: Now it's more like "update rows" instead of "add extra rows" if I have not misunderstood 🤔

CheshireSwift: The method name is "add extra row *data*". It is adding extra data to the rows, not adding rows 🙂

help_to_heat/portal/download_views.py
```diff
@@ -77,12 +77,20 @@ def add_extra_row_data(referral):
...
+    uprn = row.get("uprn")
+    uprn = '="' + str(uprn) + '"' if uprn else ""
     row = {
         **row,
+        "contact_number": contact_number,
+        "uprn": uprn,
         "ECO4": "Energy Company Obligation 4" in eligibility and "Yes" or "No",
         "GBIS": "Great British Insulation Scheme" in eligibility and "Yes" or "No",
         "epc_rating": epc_rating and epc_rating != "Not found" or "",
+        "accept_suggested_epc": accept_suggested_epc and accept_suggested_epc != "Not found" or "",
```
CheshireSwift, line 93: We should check with the client what the suppliers want here rather than assuming - "Not found" might be useful information (PC-400 was just because the date field needed to be empty when missing so that Excel autoformatted it as a date).

help_to_heat/portal/download_views.py
xnw9: Ahhh yes of cause!! :comfy-panic: :facepalm:

help_to_heat/portal/download_views.py
xnw9: Just confirmed with Ladun that we should keep "Not found" if the rest of the column won't contain numbers, so I'll remove this change!

CheshireSwift (approved):

PR #122, opened 2023-08-14
PC-NONE: Add data=None to a few get methods

JoshAdamPowell:
This has been causing 500s because, if there is an error (eg a missing field) then the base class PageView is calling the super's get function with a data variable


JoshAdamPowell:
It's almost like not having types is a mistake

AyaPK (approved):

PR #123, opened 2023-08-15
PC-427: Fix 500 error when no EPC found

AyaPK:
# PC-427: Fix 500 error when no EPC found
- Moves initial EPC calculation logic to the page before EPC checks, removing the need to overring the get method.


help_to_heat/frontdoor/views.py
```diff
@@ -301,18 +301,19 @@ def get_context(self, request, session_id, *args, **kwargs):
         return {"council_tax_band_options": council_tax_bands}
 
     def handle_post(self, request, session_id, page_name, data, is_change_page):
-        return super().handle_post(request, session_id, page_name, data, is_change_page)
+        session_data = interface.api.session.get_session(session_id)
```
AyaPK, line 304: So this logic is now happening in the handle_post for the `council tax` page. I'm admittedly not a fan of it this approach, but it means the EPC is being calculated before the council tax page redirects anywhere, so it'll redirect to the right page.

AyaPK: (I am very open to a better place to put this logic. I couldn't think of anywhere but hopefully someone else can? :D)

JoshAdamPowell:
I think nevertheless we should still merge https://github.com/UKGovernmentBEIS/help-to-heat-GBIS/pull/122 as a safety net

help_to_heat/frontdoor/views.py
```diff
@@ -301,18 +301,19 @@ def get_context(self, request, session_id, *args, **kwargs):
...
 @register_page("epc")
 class EpcView(PageView):
-    def get(self, request, session_id, page_name, data=None, errors=None, is_change_page=False):
-        if not errors:
-            errors = {}
-        data = {}
-        prev_page_url, next_page_url = get_prev_next_urls(session_id, page_name)
-        extra_context = self.get_context(request=request, session_id=session_id, page_name=page_name, data={})
-
+    def get_context(self, request, session_id, page_name, data):
```
AyaPK, line 316: Everything here is just undoing the previous PR to get it back to how it was before, so that the page properly loads and makes the necessary API calls.

CheshireSwift:
Chatted to @AyaPK to get some context on this, tl;dr okay with this approach for now but may want to shift how we do it in future. Reviewing this version of things now.

help_to_heat/frontdoor/views.py
CheshireSwift: I think the ideal version of this is for the EPC page to worry about this, and redirect if required, but as discussed that requires some additional routing/handler infrastructure. This is fine for now.

CheshireSwift (approved):
lgtm (modulo the caveats already discussed)

PR #124, opened 2023-08-15
PC-446: Add translation toggle and django translation function

chidin194:
PC-446 Translation infrastructure

PC-446: Wrap user facing strings in translation calls and add Welsh translations to .po file


chidin194:
I'm not sure whether we can change the formatting rules or disable them for these lines, but make check-python-code is failing due to the two below lines that are being deemed too long. However, since the strings are wrapped in a function call, they can't be split up over multiple lines:

image

chidin194:
Aya has found a working solution for multi-line strings (\ + Enter at desired line break), so I can implement this where we have especially long lines if you think this would be better?

chidin194:
> > \ + Enter at desired line break
> 
> If this works, it's probably the nicest option. Have you confirmed it extracts to the messages file correctly? I thought I had tried it and it didn't work, but I'd be happy to be wrong! 😄

I went through an example with Aya on the call and extraction and compilation looked good, for reference this was in schemas.py, lines 146 and 153. Initially they were being picked up by the linter as being too long, happily we discovered that this can work with extraction too :)

CheshireSwift (commented):
It looks like there's still quite a lot of `fuzzy` kicking about - I presume if we remove those, they get re-added?

chidin194:
> It looks like there's still quite a lot of `fuzzy` kicking about - I presume if we remove those, they get re-added?

@CheshireSwift the majority of these are page titles, I wrapped them but we don't have translations for titles yet - I made a list of all of them and concluded that the title text is too different for translation to be inferred. They would get re-added, but I can unwrap the page titles if you'd prefer that the fuzzies aren't present in the .po file.

CheshireSwift:
To be clear, I meant can we just remove the "fuzzy" tag from those translations 😄 No worries re: waiting for title translations, I think leaving them wrapped but untranslated for now is the best bet.

CheshireSwift:
Leaving a note here for posterity: we can use `\` for multiline strings in Python code, but not templates (which was my prior issue). Fortunately, the linter only cares about line length in Python code, not templates (so still settling for long lines there, with translation per-paragraph/-sentence to break things up).

CheshireSwift (approved):
lgtm. I know it was a laborious one, good work pushing through it 🙂

PR #130, opened 2023-08-18
PC-465: EON copy change

chidin194:
Changed string in schemas.py


help_to_heat/frontdoor/schemas.py
```diff
@@ -463,7 +463,7 @@
     "E (Gas & Electricity) Ltd",
     "Ecotricity",
     "EDF",
-    "E.ON",
+    "E.ON Next",
```
JoshAdamPowell, line 466: Do you also need to change the wording in `portal/models.py`?

chidin194: @JoshAdamPowell I was wondering about this, but it's listed as a Customer Portal ticket. Perhaps we should double check with client? @LadunOmideyiSoftwire

CheshireSwift: I take it from that now being changed that that was a yes? 🙂

CheshireSwift (approved):
👍

help_to_heat/frontdoor/schemas.py
chidin194: @CheshireSwift we discussed at GBIS stand-up and concluded that this would be changed too since we changed it in both places when we changed Octopus to Octopus Energy :)

PR #131, opened 2023-08-18
PC-461: Add feedback csv download

xnw9:
https://beisdigital.atlassian.net/browse/PC-461

image
image

Also moved `Feedback` from `frontdoor/models` to `portal/models`.


help_to_heat/portal/download_views.py
```diff
@@ -96,6 +121,23 @@ def add_extra_row_data(referral):
     return row
 
 
+def match_rows_for_feedback(feedback):
+    created_at = feedback.created_at.astimezone(london_tz)
+    page_name = feedback.page_name
+    row = dict(feedback.data)
+    row = {
+        **row,
+        "useful_for_learning": row["how-much"],
```
xnw9, line 130: Have renamed the columns so it's more obvious which question in the feedback each column corresponds to

help_to_heat/frontdoor/models.py
```diff
@@ -16,9 +16,3 @@ class Answer(utils.UUIDPrimaryKeyBase, utils.TimeStampedModel):
 
     class Meta:
         indexes = [models.Index(fields=["session_id"])]
-
-
-class Feedback(utils.UUIDPrimaryKeyBase, utils.TimeStampedModel):
-    session_id = models.UUIDField(editable=False, blank=True, null=True)
-    page_name = models.CharField(max_length=128, editable=False, blank=True, null=True)
-    data = models.JSONField(encoder=DjangoJSONEncoder, editable=False)
```
CheshireSwift, line 24: I'm not convinced it is correct to move this. I think Feedback should probably still live within the frontdoor, since that is where it's being created. (FeedbackDownload can be in portal)

If we *are* moving it, then it looks like you've created a migration to make a Feedback table in portal, but it doesn't drop the existing Feedback table from frontdoor. But as I say, I don't think we should be moving it.

help_to_heat/portal/download_views.py
```diff
@@ -59,6 +69,21 @@ def download_csv_view(request):
     return response
 
 
+@require_http_methods(["GET"])
+@decorators.requires_service_manager
+def download_feedback_view(request):
+    feedbacks = models.Feedback.objects.filter()  # always download all feedbacks for now
```
CheshireSwift, line 75: If we're not doing filtering as of yet, do we need the `.filter()`?

help_to_heat/portal/download_views.py
```diff
@@ -109,3 +151,18 @@ def create_referral_csv(referrals, file_name):
     for row in rows:
         writer.writerow(row)
     return response
+
+
+def create_feedback_csv(feedbacks, file_name):
```
CheshireSwift, line 156: Probably worth us extracting a common helper for CSV download, to avoid having to explicitly add fixes (like the BOM_UTF8 stuff) to multiple places.

help_to_heat/portal/migrations/0013_feedback_feedbackdownload.py
```diff
@@ -0,0 +1,60 @@
...
+                    models.ForeignKey(
+                        blank=True, null=True, on_delete=django.db.models.deletion.PROTECT, to=settings.AUTH_USER_MODEL
+                    ),
+                ),
+            ],
+            options={
+                "abstract": False,
+            },
+        ),
+        migrations.CreateModel(
```
CheshireSwift, line 35: As discussed, don't think we should be creating this one.

help_to_heat/portal/models.py
```diff
@@ -138,6 +138,32 @@ def __str__(self):
...
+class FeedbackDownload(utils.UUIDPrimaryKeyBase, utils.TimeStampedModel):
+    file_name = models.CharField(max_length=255, blank=True, null=True)
+    last_downloaded_by = models.ForeignKey(User, blank=True, null=True, on_delete=models.PROTECT)
+
+    @property
+    def local_created_at(self):
+        return self.created_at.astimezone(london_tz)
+
+
+class Feedback(utils.UUIDPrimaryKeyBase, utils.TimeStampedModel):
```
CheshireSwift, line 150: (Again, probably leave in frontdoor.)

help_to_heat/frontdoor/interface.py
```diff
@@ -433,7 +433,7 @@ def get_epc(self, uprn, country):
 class Feedback(Entity):
     @with_schema(load=FeedbackSchema, dump=SuccessSchema)
     def save_feedback(self, session_id, page_name, data):
-        models.Feedback.objects.create(session_id=session_id, page_name=page_name, data=data)
+        portal.models.Feedback.objects.create(session_id=session_id, page_name=page_name, data=data)
```
CheshireSwift, line 436: This is kind of the core of why I think feedback should live in the frontdoor, conceptually: the frontdoor is a less secure, more open environment. It's generally fine for anything from frontdoor to show up in portal. Portal is more secure/less public, so we shouldn't really have anything in frontdoor even referencing things in portal.

CheshireSwift (changes requested):
Main takeaway is that I don't think we should move the feedback entity to portal; it belongs in frontdoor, plus some smaller changes.

help_to_heat/frontdoor/interface.py
xnw9: Thank you for the explanation - yes that makes sense! Will change it back :)

help_to_heat/frontdoor/models.py
xnw9: If `Feedback` lives in frontdoor and `FeedbackDownload` lives in portal, I guess we still need to import `FeedbackDownload` from portal for it to be used in `Feedback`? Is this alright? :0

PR #132, opened 2023-08-18
PC-NONE: Addresses failing if none found

JoshAdamPowell:
I don't know what's changed. I swear we previously didn't 500 if there were no addresses. I wonder if the OS API's changed?


CheshireSwift:
> I wonder if the OS API's changed?

You think they might have dropped the "total results: 0" bit of the response?

CheshireSwift (approved):

JoshAdamPowell:
> > I wonder if the OS API's changed?
> 
> You think they might have swapped the empty array for omitting the results field entirely?

yeah

PR #193, opened 2023-09-11
PC-431: Exclude portal

xnw9:
image


JoshAdamPowell (approved):
This is such an unimportant change since almost every portal page is behind auth. LGTM though

PR #205, opened 2023-09-11
PC-NONE: fix cookies text

JoshAdamPowell:


CheshireSwift (approved):

CheshireSwift:
Approved, but can you raise tech debt to think about what cookie durations we actually want? Thanks.

PR #206, opened 2023-09-12
PC-384: Park home pages

xnw9:
Added three pages:
- Do you live in a park home
- Is the park home your main residence
- Not eligible because park home is not main residence

❗ `change` page logic 
Although we don't expect a lot of people changing park home answer from summary, I think we still need to consider what the flow is like if they do so.
**There is no summary line for the new pages, i.e. no way to change park home answers from summary page ATM.**

This is included in another branch park-home-with-logic 
- If they change "do you own the property" answer to "yes I own the property", they are redirected back to main flow (i.e. to `page`, instead of `change-page`)

❗ page skipping logic
A different flow for park home.
Another branch park-home-with-logic has this new flow.

❗ new tests

🎫 translation added


LadunOmideyiSoftwire:
Keep in draft til release over

xnw9:
Forgot to fix tests 🙃

CheshireSwift:
Minor point, but this is quite a lot of review; might have been nicer to split it up into three separate reviews 🙂 

(No need to change it now though, it's fine. Just something to consider in future.)

help_to_heat/frontdoor/schemas.py
```diff
@@ -168,6 +177,26 @@
...
     },
 )
+park_home_options_map = (
+    {
+        "value": "Yes",
+        "label": pgettext_lazy("park home question option", "Yes"),
+    },
+    {
+        "value": "No",
+        "label": pgettext_lazy("park home question option", "No"),
```
CheshireSwift, line 187: Interesting that these are different from other contexts.

help_to_heat/frontdoor/views.py
```diff
@@ -282,6 +286,75 @@ class OwnPropertyView(PageView):
...
+
+
+@register_page("park-home-main-residence")
+class ParkHomeMainResidenceView(PageView):
+    def get_context(self, *args, **kwargs):
+        return {"park_home_main_residence_options_map": schemas.park_home_main_residence_options_map}
+
+    def handle_post(self, request, session_id, page_name, data, is_change_page):
+        prev_page_name, next_page_name = get_prev_next_page_name(page_name)
+        data = dict(request.POST.dict())
```
CheshireSwift, line 334: Does this need the `dict(...)` call if we're already calling `.dict()`?

(Same question elsewhere we have both.)

help_to_heat/locale/cy/LC_MESSAGES/django.po
```diff
@@ -2914,3 +2974,12 @@ msgstr "Darganfyddwch a allwch chi gael cymorth gan eich cyngor lleol drwy'r "
...
 msgstr "Gronfa Cymorth i Aelwydydd"
+
+#~ msgid "2 years"
+#~ msgstr "2 flynedd"
+
+#~ msgid "24 hours"
+#~ msgstr "24 awr"
+
+#~ msgid "1 minute"
+#~ msgstr "1 funud"
```
CheshireSwift, line 2985: Are these unused? The extraction seems to think so. If so, we should remove them.

CheshireSwift (changes requested):
Couple of small things, otherwise lgtm.

help_to_heat/frontdoor/views.py
xnw9: No they don't need it indeed 😢

xnw9:
> Minor point, but this is quite a lot of review; might have been nicer to split it up into three separate reviews 🙂
> 
> (No need to change it now though, it's fine. Just something to consider in future.)

Sorry, will do! 😿

help_to_heat/frontdoor/schemas.py
```diff
@@ -54,6 +57,9 @@
     "applications-closed": {"prev": "supplier", "next": None},
     "application-closed-utility-warehouse": {"prev": "supplier", "next": None},
     "benefits": {"prev": "council-tax-band", "next": "household-income"},
+    "park-home": {"prev": "own-property", "next": "address"},
+    "park-home-main-residence": {"prev": "park-home", "next": "address"},
+    "park-home-application-closed": {"prev": "park-home-main-residence", "next": "none"},
```
davidm-m, line 62: I think that should be `None`, not `"none"`.

help_to_heat/frontdoor/schemas.py
```diff
@@ -86,6 +92,9 @@
 household_pages = {
     "country": ("country",),
     "supplier": ("supplier",),
+    "park-home": ("park_home",),
+    "park-home-main-residence": ("park_home_main_residence",),
+    "park-home-application-closed": ("park-home-application-closed",),
     "bulb-warning-page": ("bulb-warning-page",),
     "utility-warehouse-warning-page": ("utility-warehouse-warning-page",),
     "applications-closed": ("applications-closed",),
```
davidm-m, line 100: Why the inconsistency between dashes and underscores here?

help_to_heat/frontdoor/views.py
```diff
@@ -282,6 +286,75 @@ class OwnPropertyView(PageView):
...
+
+        if park_home_main_residence == "No":
+            next_page_name = "park-home-application-closed"
+
+        if is_change_page:
+            if park_home_main_residence == "No":
+                return redirect("frontdoor:page", session_id=session_id, page_name=next_page_name)
+            else:
+                assert page_name in schemas.change_page_lookup
+                next_page_name = schemas.change_page_lookup[page_name]
```
davidm-m, line 345: This looks weird to me. Why do we have the first `if`? It could be replaced by putting the `next_page_name` assignment in the other `if park_home_main_residence == "No"` block or having the return say `page_name="park-home-application-closed"`.

davidm-m: Or we could even just have
```python
if park_home_main_residence == "No":
    next_page_name = "park-home-application-closed"
elif is_change_page:
    assert page_name in schemas.change_page_lookup
    next_page_name = schemas.change_page_lookup[page_name]

return redirect("frontfoor:page", session_id=session_id, page_name=next_page_name)
```

davidm-m (changes requested):
I don't know much about django or GBIS so please push back on any of my comments, but a few things stood out to me.

help_to_heat/frontdoor/schemas.py
xnw9: You're absolutely right 😿

help_to_heat/frontdoor/schemas.py
xnw9: Ahh they should all have underscores 😅

help_to_heat/frontdoor/views.py
xnw9: Thanks for pointing out - I figured out we actually don't need `is_change_page` in this two classes in this branch 😅

help_to_heat/frontdoor/schemas.py
```diff
@@ -86,6 +92,9 @@
 household_pages = {
     "country": ("country",),
     "supplier": ("supplier",),
+    "park-home": ("park_home",),
+    "park-home-main-residence": ("park_home_main_residence",),
+    "park-home-application-closed": ("park_home_application_closed",),
```
CheshireSwift, line 97: Still got some underscores kicking about. These ones are page names, not data fields 🙂

CheshireSwift (changes requested):

help_to_heat/frontdoor/schemas.py
xnw9: Sorry I got a bit confused now :( I thought they need underscores to match those in the `summary_map`, so that we could find the corresponding summary line question? 🤔

xnw9: If I understanding is correct it should be like this as application-closed pages won't have matching summary lines 🤔 Everything else is working as before
image

CheshireSwift: Suspect you're right, just annoyed/confused by the warning pages 😓 

Fine as-is, ty

CheshireSwift (approved):

help_to_heat/frontdoor/schemas.py
xnw9: Cool! Yeah it is ready confusing 😅

PR #214, opened 2023-09-15
PC-487: Add new British Gas unavailable screen

chidin194:
image

image


tests/test_frontdoor.py
```diff
@@ -76,6 +76,29 @@ def test_flow_errors():
...
+    page = form.submit().follow()
+
+    form = page.get_form()
+    form["supplier"] = "British Gas"
+    page = form.submit().follow()
+
+    assert page.has_text("British Gas are not taking referrals right now but they will be shortly")
+
+    page = page.click(contains="Back")
+    assert page.has_one("h1:contains('Select your home energy supplier from the list below')")
```
chidin194, line 100: The NI test above checks the _page_name from the session, wondering if I should be doing that here too? I tried and it failed, couldn't work out exactly how the logic works

help_to_heat/frontdoor/views.py
```diff
@@ -84,8 +84,8 @@ def unavailable_supplier_redirect(session_id):
 
     if supplier == "Utility Warehouse":
         next_page_name = "application-closed-utility-warehouse"
-    else:
-        next_page_name = "applications-closed"
+    if supplier == "British Gas":
```
AyaPK, line 87: We'd rather use an `elif` here to keep the logic as one block rather than starting a new if statement

help_to_heat/frontdoor/views.py
```diff
@@ -539,8 +539,8 @@ def handle_post(self, request, session_id, page_name, data, is_change_page):
         request_supplier = request_data.get("supplier")
         if request_supplier == "Bulb, now part of Octopus Energy":
             next_page_name = "bulb-warning-page"
-        if request_supplier in unavailable_suppliers:
-            next_page_name = "applications-closed"
+        if request_supplier == "British Gas":
```
AyaPK, line 542: Is there a reason we totally remove the old `request_supplier in unavailable_suppliers` logic to have something unique to british gas?

tests/test_frontdoor.py
AyaPK: I think the test as you've written it does the job fine

AyaPK (commented):
A few things to note!

help_to_heat/frontdoor/views.py
chidin194: Will change this, thanks!

help_to_heat/frontdoor/views.py
chidin194: I think it's going to be redundant logic at the moment since British Gas and Utility Warehouse are the only unavailable suppliers and so we're not making use of the applications-closed page, although if you'd rather we kept it in as a catch-all should we add more unavailable suppliers in the future, I can add it back in!

AyaPK: No, I agree with you, if we don't use it at any point it should be removed! (YAGNI)

chidin194: Great, sounds good, thanks!

CheshireSwift: I'd actually go the opposite way on this: proliferating extra views for every supplier that's unavailable isn't great. The plan was actually already to move Utility Warehouse to use the same view as the "standard" one we were using for British Gas, with that view being slightly more dynamic and some logic to vary what is shown as required.

I'd still rather we take that approach here: one template that changes the content based on the supplier, rather than distinct views for each.

CheshireSwift (changes requested):

CheshireSwift (approved):

PR #218, opened 2023-09-22
PC-500: fix address bug

JoshAdamPowell:
The code here has very "copied off stack overflow" vibes so I'm not entirely sure if I'm breaking something here. But it fixes the bug.

### DIFF
--- a/help_to_heat/frontdoor/interface.py
+++ b/help_to_heat/frontdoor/interface.py
@@ -378,9 +378,7 @@ class Address(Entity):
 
         if "DEPENDENT_THOROUGHFARE_NAME" and "THOROUGHFARE_NAME" in dpa:
             line_1_parts.append(dpa["THOROUGHFARE_NAME"].title())
-        else:
             line_1_parts.append(dpa["DEPENDENT_THOROUGHFARE_NAME"].title())
-            line_2_parts.insert(0, dpa["THOROUGHFARE_NAME"].title())
 
         line1 = ", ".join(line for line in line_1_parts if line)
         line2 = ", ".join(line for line in line_2_parts if line)


CheshireSwift:
It's replicating this logic: https://github.com/UKGovernmentBEIS/desnz-home-energy-retrofit-beta/blob/develop/HerPublicWebsite.BusinessLogic/ExternalServices/OsPlaces/OsPlacesDpaDto.cs#L27

If the behaviour is incorrect, we should think about whether it's also an issue in HUG2.

help_to_heat/frontdoor/interface.py
```diff
@@ -376,8 +376,9 @@ def parse_dpa_to_address(self, dpa):
...
             line_1_parts.append(dpa["DEPENDENT_LOCALITY"].title())
 
-        if "DEPENDENT_THOROUGHFARE_NAME" and "THOROUGHFARE_NAME" in dpa:
-            line_1_parts.append(dpa["THOROUGHFARE_NAME"].title())
+        if "DEPENDENT_THOROUGHFARE_NAME" not in dpa:
+            if "THOROUGHFARE_NAME" in dpa:
+                line_1_parts.append(dpa["THOROUGHFARE_NAME"].title())
         else:
             line_1_parts.append(dpa["DEPENDENT_THOROUGHFARE_NAME"].title())
             line_2_parts.insert(0, dpa["THOROUGHFARE_NAME"].title())
```
CheshireSwift, line 384: Theoretically fails if we have dependent but not thoroughfare?

JoshAdamPowell: indeed so. Sorry I feel like I've been crap fixing this bug

CheshireSwift (approved):

PR #220, opened 2023-09-26
PC-501: Handle unanswered multiple-choice questions on the feedback form to fix downloads

chidin194:
Added "-" as a fallback in case of a key error/unanswered question for any of the multiple-choice questions, which will be shown in the CSV. The long form text question, if unanswered, just shows as "" in the CSV

### DIFF
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -203,9 +203,9 @@ def match_rows_for_feedback(feedback):
     row = dict(feedback.data)
     row = {
         **row,
-        "useful_for_learning": row["how-much"],
-        "sufficient_guidance": row["guidance-detail"],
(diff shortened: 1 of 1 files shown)


CheshireSwift:
Is there a reason we're using `"-"` rather than just `""`? Otherwise, fbm.

chidin194:
> Is there a reason we're using `"-"` rather than just `""`? Otherwise, fbm.

Changed from "-" to "unanswered" as, following discussion, decided this was a better choice since the CSV file will be read by portal users and not computers

CheshireSwift (changes requested):
Minor, but if we're going with "unanswered", can it be capitalised as "Unanswered"? 🙂

CheshireSwift (approved):

PR #223, opened 2023-09-27
PC-502: Allow service managers to download referrals

AyaPK:
- Adds two buttons to the suppliers page
  - One to download all referrals
  - One to download referrals made in the last week
- They download in excel format
- They omit all customer PII

### DIFF
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -9,6 +9,7 @@ from django.utils import timezone
 from django.views.decorators.http import require_http_methods
 
 from help_to_heat.frontdoor import models as frontdoor_models
+from help_to_heat.portal import models as portal_models
 from help_to_heat.frontdoor.eligibility import calculate_eligibility
 from help_to_heat.portal import decorators, models
 
@@ -28,6 +29,30 @@ referral_column_headings = (
     "address_line_1",
     "postcode",
     "address",
+    "council_tax_band",
+    "property_type",
+    "property_subtype",
+    "epc_rating",
+    "accept_suggested_epc",
+    "epc_date",
+    "number_of_bedrooms",
+    "wall_type",
+    "wall_insulation",
+    "loft",
+    "loft_access",
+    "loft_insulation",
+    "Property main heat source",
+    "supplier",
+    "submission_date",
+    "submission_time",
+)
+
+referral_column_headers_no_pii = (
+    "ECO4",
+    "GBIS",
+    "own_property",
+    "benefits",
+    "household_income",
     "council_tax_band",
     "property_type",
     "property_subtype",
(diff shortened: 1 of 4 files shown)


tests/test_frontdoor.py
```diff
@@ -100,6 +100,7 @@ def test_british_gas_unavailable():
     page = page.click(contains="Back")
     assert page.has_one("h1:contains('Select your home energy supplier from the list below')")
 
+
```
AyaPK, line 103: idk the linter did this

help_to_heat/portal/download_views.py
```diff
@@ -78,6 +105,38 @@ def create_referral_csv(referrals, file_name):
     return response
 
 
+def create_referral_all_xlsx(referrals, file_name):
```
AyaPK, line 108: in hindsight this is probably better named as something like `create_referral_xlsx_no_pii`

help_to_heat/portal/download_views.py
```diff
@@ -138,6 +197,26 @@ def download_feedback_view(request):
...
+    downloaded_at = timezone.now()
+    file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
+    response = create_referral_all_xlsx(feedbacks, file_name)
+    return response
+
+
+@require_http_methods(["GET"])
+@decorators.requires_service_manager
+def download_referrals_last_week_view(request):
+    feedbacks = portal_models.Referral.objects.filter(created_at__gte=datetime.now() - timedelta(days=7))
```
AyaPK, line 213: I only discovered this syntax today and it's SO COOL!?

CheshireSwift: The filter syntax? It's pretty good, yeah :D

AyaPK: Oh specifically the date filtering here!

CheshireSwift: Yeah, time deltas are fun.

help_to_heat/portal/download_views.py
```diff
@@ -138,6 +197,26 @@ def download_feedback_view(request):
     return response
 
 
+@require_http_methods(["GET"])
+@decorators.requires_service_manager
+def download_referrals_all_view(request):
+    feedbacks = portal_models.Referral.objects.all()
```
chidin194, line 203: Was just looking at the PR for general learning, but I think this variable might want to be named `referrals` as opposed to `feedbacks` in this function and the one below?

davidm-m (changes requested):
Mostly looks good, I think we've got some good opportunities for commonisation though!

davidm-m (approved):
Amazing! Those methods are looking much nicer.
One question before you merge: what makes these methods wet??

AyaPK:
> Amazing! Those methods are looking much nicer. One question before you merge: what makes these methods wet??

I wouldn't say that they are anymore, but previously we had methods that basically did the same thing and a bunch of repeated code. (Admittedly I might be using terminology wrong in this commit!)

PR #236, opened 2023-10-18
PC-489: Change supplier portal session duration to 15 minutes

chidin194:
Extends session duration from 10 to 15 minutes


CheshireSwift (approved):
I'm a bit surprised we're doing 10 => 15; feels kinda insignificant. But if that's what people are happy with then change looks fine.

chidin194:
> I'm a bit surprised we're doing 10 => 15; feels kinda insignificant. But if that's what people are happy with then change looks fine.

Yeah, I think a few complaints around session duration were received from suppliers around being logged out, but this is the maximum we can increase it to while still adhering to national cyber-security guidelines.

PR #239, opened 2023-10-20
PC-505: Add new GBIS feedback questions

chidin194:
New GBIS feedback form, pending Welsh translations from DESNZ


chidin194:
From a design perspective, is this error enough? I was also considering adding a red margin to each question, but that might be confusing since none are specifically mandatory:

image

help_to_heat/frontdoor/views.py
```diff
@@ -725,24 +725,41 @@ def get_context(self, session_id, *args, **kwargs):
...
+            if session_id and page_name:
+                return redirect("frontdoor:feedback-thanks", session_id=session_id, page_name=page_name)
+
+            else:
+                return redirect("frontdoor:feedback-thanks")
+        
+    def validate(self, data):
+        if (not data.get("satisfaction")) and (not data.get("usage-reason")) and (not data.get("guidance"))  and (not data.get("accuracy")) and (not data.get("advice")) and (data.get("more-detail") == ''):
+            errors = {
+                "data": "Please answer at least one question"
```
chidin194, line 758: This value is never actually used. The `errors` variable is only checked for the presence of errors, and then, depending on whether it's truthy/falsy, the div containing the error alert on the page is rendered. Perhaps I should change this string to a boolean to avoid confusion?

help_to_heat/templates/frontdoor/feedback.html
```diff
@@ -2,7 +2,7 @@
 {% import "macros.html" as macros with context %}
 
 {% block title %}
-  <title>{{_("Give us some feedback - Check eligibility for Great British Insulation Scheme - GOV.UK")}}</title>
+  <title>{{_("Help us improve the service - Check eligibility for Great British Insulation Scheme - GOV.UK")}}</title>
```
chidin194, line 5: Changing as I thought slightly more delicate wording would be more appropriate 😅

help_to_heat/templates/frontdoor/feedback.html
```diff
@@ -12,72 +12,103 @@
...
 
+    {% if errors %}
+    <div class="govuk-error-summary" data-module="govuk-error-summary">
+      <div role="alert">
+        <h2 class="govuk-error-summary__title">
+          {{_("Please answer at least one question before submitting feedback")}}
+        </h2>
+      </div>
+    </div>
+    {% endif %}
```
chidin194, line 26: This is rendered conditionally and is adapted from the error summary div in `macros.html`

chidin194:
New GBIS feedback form, pending Welsh translations from DESNZ

help_to_heat/frontdoor/views.py
AyaPK: I'd agree with that, if this summarisation is correct

help_to_heat/templates/frontdoor/feedback.html
AyaPK: As long as the client is happy with this change :D

AyaPK (commented):
A few things to look at but mostly looks good!

chidin194:
> Regarding the error message
> 
> > From a design perspective, is this error enough? I was also considering adding a red margin to each question, but that might be confusing since none are specifically mandatory:
> > image
> 
> I'd say to look at the component style guidelines, and actually use an error summary component for this https://design-system.service.gov.uk/components/error-summary/

So I've used the error summary class, but this way of validating a form doesn't fit in with the usual validation pattern on the design system as we're not validating any one specific field in particular. Maybe this needs a separate conversation with the client!

AyaPK:
> So I've used the error summary class, but this way of validating a form doesn't fit in with the usual validation pattern on the design system as we're not validating any one specific field in particular. Also, adding a separate error message alongside each field might require more extensive changes since we get the data in a blob. Maybe this needs a separate conversation with the client! @AyaPK @CheshireSwift

I'd posit you could do just do something like:
```
<div class="govuk-error-summary" data-module="govuk-error-summary">
  <div role="alert">
    <h2 class="govuk-error-summary__title">
      There was a problem submitting the form
    </h2>
    <div class="govuk-error-summary__body">
      <ul class="govuk-list govuk-error-summary__list">
        <li>
          <a href="#">Please answer at least one question before submitting feedback</a>
        </li>
      </ul>
    </div>
  </div>
</div>
```
which would look much nicer, fitting the style guidelines and using the component as intended

help_to_heat/frontdoor/views.py
chidin194: Changed this, thanks!

help_to_heat/templates/frontdoor/feedback.html
chidin194: What do you think, @CheshireSwift? I don't think there have been any specific requirements around titles, iirc?

chidin194:
New and improved error screen:
image

help_to_heat/frontdoor/schemas.py
```diff
@@ -598,34 +598,81 @@
         "value": "No loft",
     },
 )
-multichoice_options = (
+multichoice_options_agreement = (
```
CheshireSwift, line 601: Minor, but I think it's better to start the name with the most pertinent information. In this case, I think `satisfaction_multichoice_options`, `agreement_multichoice_options` and `service_usage_multichoice_options` make intent/distinctions a bit clearer.

help_to_heat/frontdoor/views.py
```diff
@@ -725,24 +725,39 @@ def get_context(self, session_id, *args, **kwargs):
...
         }
         return render(request, template_name=template_name, context=context)
 
     def post(self, request, session_id=None, page_name=None):
         data = request.POST.dict()
-        interface.api.feedback.save_feedback(session_id, page_name, data)
-        if session_id and page_name:
-            return redirect("frontdoor:feedback-thanks", session_id=session_id, page_name=page_name)
+        errors = self.validate(data)
+        if errors is True:
```
CheshireSwift, line 745: Can this just be `if errors:`?

help_to_heat/templates/frontdoor/feedback.html
CheshireSwift: Hmm. I thought there was a standardised title format beta feedback pages, but it doesn't look like it. Still, it's user-facing content, so I'd want to check it with stakeholders, yeah.

help_to_heat/templates/frontdoor/feedback.html
```diff
@@ -12,72 +12,110 @@
...
+          </ul>
+        </div>
+      </div>
+    </div>
+    {% endif %}
+
+    <div class="govuk-form-group govuk-!-padding-bottom-7">
+      <fieldset class="govuk-fieldset">
+        <legend class="govuk-fieldset__legend govuk-fieldset__legend--l">
+          <h1 class="govuk-label-wrapper">
```
CheshireSwift, line 38: Looks like yours isn't the original sin here, but we shouldn't have multiple `h1` tags in a page. All these questions should probably be h2.

CheshireSwift (changes requested):
Couple of things, but mostly good.

help_to_heat/frontdoor/views.py
chidin194: Changed 😄

help_to_heat/templates/frontdoor/feedback.html
chidin194: Got it, thanks, changed the questions to h2!

help_to_heat/frontdoor/schemas.py
chidin194: Changed these variable names 😄

help_to_heat/templates/frontdoor/feedback.html
chidin194: If we think the change is worth it I'll ask, otherwise I can just leave it as is? @CheshireSwift

CheshireSwift: If they want it out ASAP I'd be tempted to just leave it as it was 🙂

CheshireSwift (approved):

PR #243, opened 2023-11-01
PC-533: Update ECO4 eligibility logic

chidin194:
Updated eligibility.py and accompanying tests and tested all cases manually


help_to_heat/frontdoor/eligibility.py
```diff
@@ -30,25 +28,34 @@ def calculate_eligibility(session_data):
...
+    
+    # ECO4 and GBIS scenario 1 - home owner
     if country in country_council_tax_bands:
-        if council_tax_band in country_council_tax_bands[country]["eligible"]:
-            if epc_rating in ("E", "F", "G", "Not found"):
+        if own_property in ("Yes, I own my property and live in it",):
+            if epc_rating in ("D","E", "F", "G", "Not found"):
                 if benefits in ("Yes",):
-                    logger.error("Scenario 1")
+                    logger.error("ECO4 scenario 1")
```
chidin194, line 42: Did we want to remove this logging?

help_to_heat/frontdoor/eligibility.py
```diff
@@ -30,25 +28,34 @@ def calculate_eligibility(session_data):
...
             if epc_rating in ("E", "F", "G", "Not found"):
                 if benefits in ("Yes",):
-                    logger.error("Scenario 2")
+                    logger.error("ECO4 scenario 2")
+                    return ("GBIS", "ECO4")
+                
+    # ECO4 and GBIS scenario 3 - social housing tenant
+    if country in country_council_tax_bands:
+        if own_property in ("No, I am a social housing tenant",):
+            if epc_rating in ("D", "E", "F", "G", "Not found"):
```
chidin194, line 56: I think scenarios 1 and 3 could be combined into a single scenario as the requirements in terms of EPC and benefits are the same, it would just be a case of combining the logic in lines 55 and 39? I've laid it out into 3 scenarios here as that's how Ian communicated the ECO4 eligibility to us.

help_to_heat/frontdoor/eligibility.py
CheshireSwift: Happy either way. They should be released up through the envs together, so we definitely want to do it before releasing the changes to UAT.

Do you have thoughts on whether we should remove them or switch them to log at debug level? I'd lean towards remove (since they can be re-added as required if seeing scenario selection is useful), but you've been more hands-on with dev than me.

help_to_heat/frontdoor/eligibility.py
CheshireSwift: In this one instance, I'd be inclined to keep the logic matching our source of logic. Since the eligibility is a bit knotty (and potentially liable to change), I'm keen to have as little translation between requirements and code as possible.

CheshireSwift (approved):
Code looks fine and logic matches ticket spec.

I just want to confirm that in the scenarios where we're no longer considering council tax bands, that is intended? i.e. these are intended to replace those scenarios.

As an aside, it'd be really good if we could get a cleaner way of receiving scenario specs, specifically one that translates nicely into unit tests. If we can unit test the scenarios against the spec with confidence, then we can refactor the eligibility logic without worrying we're introducing bugs.

help_to_heat/frontdoor/eligibility.py
chidin194: I think they can be removed, I did quite a lot of manual testing and the logging proved useful, but I'm happy that it's served its purpose and can be removed

help_to_heat/frontdoor/eligibility.py
chidin194: Sounds good, thanks!

PR #246, opened 2023-11-13
PC-559: Add BG back

chidin194:
Removed British Gas from the unavailable suppliers list, but haven't removed the logic in case we'll need it in the future. If that's unlikely, I can amend the PR to remove it entirely @LadunOmideyiSoftwire


LadunOmideyiSoftwire (approved):
Yes, 100% keep the logic. We'll probably need to turn suppliers on and off for ever.

PR #249, opened 2023-12-01
PC-519: CSV changes for service managers

chidin194:
Additions to the CSV requested by DESNZ, in particular:

- Country
- Postcode
- EPC letter rating (in the case where we don't find a rating, I've kept the original logic intact, so we will still populate "", instead of "Not found" - hopefully this is fine

What the CSV looks like now locally for me:
image


LadunOmideyiSoftwire (approved):
LGTM!

PR #253, opened 2023-12-19
PC-580: Change feedback form link and fix test

xnw9:
https://beisdigital.atlassian.net/browse/PC-580

### DIFF
--- a/help_to_heat/templates/frontdoor/base.html
+++ b/help_to_heat/templates/frontdoor/base.html
@@ -99,7 +99,7 @@
         </strong>
         <span class="govuk-phase-banner__text">
           {{_("This is a new service – your ")}}<a class="govuk-link"
-            href="{% if session_id and page_name %}{{url('frontdoor:feedback', session_id=session_id, page_name=page_name)}}{% else %}{{url('frontdoor:feedback')}}{% endif %}"
+            href="https://forms.office.com/Pages/ResponsePage.aspx?id=BXCsy8EC60O0l-ZJLRst2JbIaLaL_oJOivXjaXYvCetUMFRBRVcyWkk4TzhYU0E4NjQzWlZMWThRMC4u"
             tabindex="0">{{_("feedback")}}</a>{{_(" will help us to improve it.")}}
         </span>
       </p>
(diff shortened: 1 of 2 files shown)


tests/test_frontdoor.py
```diff
@@ -682,31 +663,7 @@ def test_feedback_with_session():
...
-    assert feedback.data["usage-reason"] == "To find ways to reduce my energy bills"
-    assert feedback.data["guidance"] == "Agree"
-    assert feedback.data["accuracy"] == "Neutral"
-    assert feedback.data["advice"] == "Agree"
-    assert feedback.data["more-detail"] == "Improvement comment"
-
-    feedback_session_id = page.path.split("/")[3]
-    assert uuid.UUID(feedback_session_id)
-
-    page = page.click(contains="Return to your application")
```
xnw9, line 708: There doesn't seem to be a way to go back to previous page in this case as it doesn't maintain a history of visited page?

If it is the case then we might just delete these tests?

AyaPK: Yep, since we don't look after this page we can't add one either, purge the test :D

help_to_heat/templates/frontdoor/base.html
```diff
@@ -99,7 +99,7 @@ <h2 class="govuk-cookie-banner__heading govuk-heading-m">{{_("Cookies we are usi
         </strong>
         <span class="govuk-phase-banner__text">
           {{_("This is a new service – your ")}}<a class="govuk-link"
-            href="{% if session_id and page_name %}{{url('frontdoor:feedback', session_id=session_id, page_name=page_name)}}{% else %}{{url('frontdoor:feedback')}}{% endif %}"
+            href="https://forms.office.com/Pages/ResponsePage.aspx?id=BXCsy8EC60O0l-ZJLRst2JbIaLaL_oJOivXjaXYvCetUMFRBRVcyWkk4TzhYU0E4NjQzWlZMWThRMC4u"
```
AyaPK, line 102: When testing accessing this URL, it seems to give a permission error.
Does this require action from DESNZ or could we possibly have the wrong link?

xnw9: It's on their side; it's supposed to be like this for now :D

AyaPK: Awesome!

AyaPK:
LGTM!

AyaPK (approved):

PR #259, opened 2024-01-08
PC-570 wording changes following supplier feedback

samyou-softwire:
closes https://beisdigital.atlassian.net/browse/PC-570

a series of small wording changes & addition of extra paragraphs where needed

an example is given
old:
image

new:
image

new welsh:
image


help_to_heat/locale/cy/LC_MESSAGES/django.po
jamiehumphries, line 1: Lots of the Welsh mixes between using `'` and `’` for apostrophes. We should be using `’`.

Please can you replace these everywhere? We could take the opportunity to do it in all our strings.

jamiehumphries (changes requested):
There are a couple of bits to change based on compliance with GOV.UK style. Otherwise, this looks good.

PR #261, opened 2024-01-12
PC-607 new tickbox on confirm and submit page

samyou-softwire:
closes https://beisdigital.atlassian.net/browse/PC-607

since the check box response isn't sent to the database anyway there wasn't too much major change needed for this

does restructure the page slightly to move the information text to not be on the same padding as the checkboxes, since it isn't a check box


jamiehumphries (changes requested):
The inline errors are not displaying correctly.

They should look something like this:
image

but are displaying as:
image

In particular, they should not be links.

samyou-softwire:
@jamiehumphries 
image

jamiehumphries (changes requested):
At the moment, if you submit with only one of the boxes checked, both of them are cleared when the page reloads.

Please can you make a fix so that it remembers checked boxes?

samyou-softwire:
@jamiehumphries should now be the case
image

jamiehumphries (approved):
Looks good, thanks! Go ahead and merge. 👍

PR #263, opened 2024-01-22
PC-584 PC-514: unique identifier for each referral

samyou-softwire:
closes PC-584 & PC-514 (since they are so similar)

adds a new migration that adds a postgres SERIAL column to the database which assigns a unique referral id to each referral. this is true retroactively, existing referrals will be given a referral id

this referral id is shown at form completion
english:
image

welsh:
image

as well as included in emails
english:
```
Subject: Referral to British Gas successful
From: test@example.com
To: samuel.young@softwire.com
Date: Mon, 22 Jan 2024 14:34:49 -0000
Message-ID: <170593408992.123.6954768841899672035@75d2c0b4dff6>

Hi,

Your reference number is GBIS0000168

Your details have been submitted to British Gas.

...
```

welsh:
```
Subject: Cwblhau atgyfeiriad
From: test@example.com
To: email@example.com
Date: Mon, 22 Jan 2024 17:18:23 -0000
Message-ID: <170594390356.123.15128990575147716244@19e53f73ebac>

Shwmae,

Eich rhif cyfeirnod yw GBIS0000166

Mae’ch manylion chi wedi'u cyflwyno i British Gas.

...
```

this referral id is then exported in the csv
image
the csv is now ordered by referral id. this is not specified in the ticket, but seemed the most sensible way to order things (previously no ordering was applied)


CheshireSwift:
Only potential concern I can see is security audit complaining about publicly visible serial identifiers. They're not used as inputs anywhere though, so it should be fine? And I think something like GUIDs would be sufficiently user-unfriendly as to not be worth it.

CheshireSwift (approved):

PR #266, opened 2024-02-05
PC-708: Adjust weekly feedback download date range to one clear week

chidin194:
There are some formatting changes included here that aren't related to my changes but have appeared after running `make check-python-code`

### DIFF
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -1,7 +1,7 @@
 import codecs
 import csv
 import io
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 
 import xlsxwriter
 from dateutil import tz
@@ -154,7 +154,9 @@ def create_referral_xlsx(referrals, file_name, exclude_pii=False):
 
 
 def handle_create_spreadsheet_request(request, creator):
-    referrals = models.Referral.objects.filter(referral_download=None, supplier=request.user.supplier).order_by("referral_id")
+    referrals = models.Referral.objects.filter(referral_download=None, supplier=request.user.supplier).order_by(
+        "referral_id"
+    )
     downloaded_at = timezone.now()
     file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
     new_referral_download = models.ReferralDownload.objects.create(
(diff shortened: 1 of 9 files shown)


help_to_heat/portal/download_views.py
```diff
@@ -181,22 +183,16 @@ def download_feedback_view(request):
...
-    referrals = portal_models.Referral.objects.filter(created_at__gte=datetime.now() - timedelta(days=7)).order_by("referral_id")
-    downloaded_at = timezone.now()
-    file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
+    today = date.today()
+    end_of_last_week = today - timedelta(days=today.weekday())
+    start_of_last_week = end_of_last_week - timedelta(days=7)
+    referrals = portal_models.Referral.objects.filter(
+        created_at__gte=start_of_last_week, created_at__lt=end_of_last_week
+    ).order_by("referral_id")
+    file_name = f"Weekly Referrals ({datetime.strptime(str(start_of_last_week), '%Y-%m-%d').strftime('%d-%m-%Y')})"
```
chidin194, line 195: This is perhaps a little messy, but it seems there's no method for just formatting a date, so it seems we have to parse the date start_of_last_week as a string to a datetime, and then format this back into a string with just the date.

jamiehumphries:
> There are some formatting changes included here that aren't related to my changes but have appeared after running `make check-python-code`

Thanks for clarifying @chidin194.

No need to change this time, but in future it would be helpful if such changes could be separated out into a separate pull request, or at least a separate commit within this one.

Separating them out makes a clear distinction between functional code changes and style ones.

help_to_heat/portal/download_views.py
jamiehumphries: Are you sure about this? I was able to just run:
```
start_of_last_week.strftime("%d-%m-%Y")
```

Here it is in the docs:
https://docs.python.org/3/library/datetime.html#datetime.date.strftime

> `date.strftime(format)`
Return a string representing the date, controlled by an explicit format string. Format codes referring to hours, minutes or seconds will see 0 values. See also [strftime() and strptime() Behavior](https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior) and [date.isoformat()](https://docs.python.org/3/library/datetime.html#datetime.date.isoformat).

help_to_heat/portal/download_views.py
```diff
@@ -181,22 +183,16 @@ def download_feedback_view(request):
...
-    return response
-
-
 @require_http_methods(["GET"])
 @decorators.requires_service_manager
 def download_referrals_last_week_view(request):
-    referrals = portal_models.Referral.objects.filter(created_at__gte=datetime.now() - timedelta(days=7)).order_by("referral_id")
-    downloaded_at = timezone.now()
-    file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
+    today = date.today()
```
jamiehumphries, line 189: Please can you add a comment to explain the boundaries, like:
```
# Weekly boundaries are Mondays at midnight (00:00).
today = …
end_of_last_week = …
start_of_last_week = …
```

It's not immediately obvious from the code, so good to help out future devs.

help_to_heat/portal/download_views.py
```diff
@@ -181,22 +183,16 @@ def download_feedback_view(request):
...
+    end_of_last_week = today - timedelta(days=today.weekday())
+    start_of_last_week = end_of_last_week - timedelta(days=7)
```
jamiehumphries, line 191: I think we can actually use:
```
- timedelta(weeks=1)
```
https://docs.python.org/3/library/datetime.html#timedelta-objects

jamiehumphries (changes requested):

help_to_heat/portal/download_views.py
chidin194: Ah, thank you! It didn't seem quite right. I've changed this 😄

help_to_heat/portal/download_views.py
chidin194: Added!

help_to_heat/portal/download_views.py
chidin194: Changed this! Is the advantage here increased readability?

jamiehumphries: Yes, just a style thing. We may as well use it, since it's available.

jamiehumphries (approved):
Looks good! Let's go ahead and merge.

PR #267, opened 2024-02-05
PC-691: Removing So Energy from the user journey

samyou-softwire:
SO energy is shallowly removed from the suppliers page

they are left in the schema for compatibility


jamiehumphries:
**⚠️ Not approved.**
We should fix this properly with a database flag. Not with a hardcoded list.

PR #270, opened 2024-02-06
PC-691: Removing So Energy from the user journey

samyou-softwire:
removes so energy from the schema

the supplier page is generated from the schema and so so energy is removed from there
image

attempting to edit a radio button to so energy will no longer pass validation
image
image

though users with access to so energy can still view old records
image

since the portal tracks suppliers in a database user management is unaffected by this change
image


jamiehumphries:
Thanks for writing up all the testing detail @samyou-softwire!

Please go ahead and create a Staging pull request as well.

PR #278, opened 2024-02-22
PC-867: Update eventlet

jdgage:
This PR simply updates eventlet in-place. Have tested the public site with `DEBUG=False` and all seems fine. Haven't tested the portal - I think there's some effort in setting up locally.


jamiehumphries (commented):
Looks good, thanks. We will need to liaise with BEIS Digital / Databarracks about retrying the release to the Dev environment, so hold off on merging this PR for now.

jamiehumphries (approved):

PR #285, opened 2024-03-15
PC-923 Create "Shell warning page" and amend customer journey for Shell customers to Shell/Octopus journey

samyou-softwire:
closes [PC-923](https://beisdigital.atlassian.net/browse/PC-923)

adds a new page for shell warning, based on the similar bulb warning page

for the class in the views I left it empty, unlike the bulb warning page class. the `get_context` definition has no impact, since the templates for all the warning pages do not use any context

translations contain a few leftovers from the previously merged PR

back button functionality is the same as the other warning pages, as is functionality with the change page

contains a verifying test

image
image
image


help_to_heat/frontdoor/views.py
```diff
@@ -797,6 +799,11 @@ def get_context(self, session_id, *args, **kwargs):
         return {"supplier": supplier}
 
 
+@register_page("shell-warning-page")
+class ShellWarningPageView(PageView):
+    pass
```
jdgage, line 804: Why's this just `pass`, rather than getting the supplier like the others?

jdgage (dismissed):

help_to_heat/frontdoor/views.py
samyou-softwire: its weird - I can't find any particular reason why the above classes put the supplier into context (since the pages they render are static anyway), so imo we should consolidate one way or the other

jdgage (approved):

PR #286, opened 2024-03-18
PC-904 Include Shell referrals in Octopus csv

samyou-softwire:
closes [PC-904](https://beisdigital.atlassian.net/browse/PC-904)

though the ticket implies this is backend work, the actual substitution takes place on the frontend

follows the same format as the other converter logic, where on submission the 'supplier' value is changed

note the 'user_selected_supplier' field on the csv will always show the supplier selected, fulfilling the requirement to be able to differentiate pure octopus referrals

includes a new test to verify


tests/test_frontdoor.py
```diff
@@ -982,3 +984,17 @@ def test_utility_warehouse_to_eon():
     referral = models.Referral.objects.get(session_id=session_id)
     assert referral.supplier.name == "E.ON Next"
     referral.delete()
+
+
+@unittest.mock.patch("help_to_heat.frontdoor.interface.EPCApi", MockEPCApi)
+def test_shell_to_octopus():
```
jdgage, line 990: Not going to suggest we attack existing tests here, but at least when adding new ones I think we should try for good naming: can this be more descriptive of what it's actually testing? Something of the form "X should Y [when Z]"?

jdgage (changes requested):
One thing on the ticket that I want to check is that we have the requirement that Shell referrals should have "Shell" noted against them somewhere in the export CSV. Does that already happen automatically, or is it missing?

samyou-softwire:
> One thing on the ticket that I want to check is that we have the requirement that Shell referrals should have "Shell" noted against them somewhere in the export CSV. Does that already happen automatically, or is it missing?

@jdgage yes - the export csv always has a 'user selected supplier' field which is always what the user selects on the supplier select page

jdgage (approved):

PR #287, opened 2024-03-20
PC-912 Admin URL to extract all referrals

samyou-softwire:
to help out with [PC-912](https://beisdigital.atlassian.net/browse/PC-912)

adds a new url to the site `/portal/download-all`

this link will trigger a download of all referrals in the database

there may be issues with this potentially failing on extracting referrals in an old format (as questions asked on the site have changed over time), so this will need to be monitored

### DIFF
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -329,7 +329,8 @@ def create_feedback_csv(feedbacks, file_name):
 
 
 @require_http_methods(["GET"])
-def download_all_referrals(request):
+@decorators.requires_service_manager
+def download_all_referrals(_request):
     referrals = models.Referral.objects.all()
     downloaded_at = timezone.now()
     file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -326,3 +326,12 @@ def create_feedback_csv(feedbacks, file_name):
     full_file_name = f"feedback-data-{file_name}"
     response = create_csv(feedback_column_headings, rows, full_file_name)
     return response
+
+
+@require_http_methods(["GET"])
+def download_all_referrals(request):
+    referrals = models.Referral.objects.all()
+    downloaded_at = timezone.now()
+    file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
+    response = create_referral_csv(referrals, file_name)
+    return response
--- a/help_to_heat/portal/urls.py
+++ b/help_to_heat/portal/urls.py
@@ -101,6 +101,7 @@ portal_patterns = [
     path("user/<uuid:user_id>/verify-otp/<str:token>/", authentication_views.VerifyOTPView, name="verify-otp"),
     path("accounts/login/", authentication_views.CustomLoginView, name="account_login"),
     path("accounts/", include("allauth.urls")),
+    path("download-all/", download_views.download_all_referrals, name="download_all_referrals"),
 ]
 
 if SHOW_EPC_UPLOADS_PAGE:


help_to_heat/portal/download_views.py
```diff
@@ -326,3 +326,13 @@ def create_feedback_csv(feedbacks, file_name):
...
     return response
+
+
+@require_http_methods(["GET"])
+@decorators.requires_service_manager
+def download_all_referrals(_request):
+    referrals = models.Referral.objects.all()
+    downloaded_at = timezone.now()
+    file_name = downloaded_at.strftime("%d-%m-%Y %H_%M")
+    response = create_referral_csv(referrals, file_name)
```
jdgage, line 337: From a quick look, I think this will blow up when any of `contact_number`, `country`, `epc_rating`, or `council_tax_band` is missing from `data`. There's some referrals in the prod DB with the last two missing, so I think we'll need to make sure our code is robust to this.

I think we can do that by adding defaults to the `dict.get` calls, so it doesn't default to `None`. Would you mind looking at doing that?

Please also double-check my working in getting those columns, and maybe have a look to see if there's any such referrals in dev or uat that we can test with?

help_to_heat/portal/urls.py
```diff
@@ -101,6 +101,7 @@
     path("user/<uuid:user_id>/verify-otp/<str:token>/", authentication_views.VerifyOTPView, name="verify-otp"),
     path("accounts/login/", authentication_views.CustomLoginView, name="account_login"),
     path("accounts/", include("allauth.urls")),
+    path("download-all/", download_views.download_all_referrals, name="download_all_referrals"),
```
jdgage, line 104: `download-all-referrals` maybe?

jdgage (changes requested):
Would you mind also putting up a PR to remove this, so we can merge it after release?

help_to_heat/portal/download_views.py
samyou-softwire: looking into this, I wasn't sure what suggests country or council_tax_band are necessary fields? deleting them from my local data seems to work okay, & I was under the impression only fields with logic applied (in this case string concatenation) in `add_extra_row_data` were pertinent

added defaults for `contact_number` & `uprn` by this logic, seems to export the csvs from the admin command fine enough

jdgage: I might be wrong - I expect you've got more Python experience than me - but I thought that evaluating `None in ("tuple", "of", "strings")` would error? We (potentially) do that with `country`, `epc_rating`, and `council_tax_band` inside `calculate_eligibility`. It might be that we never hit the problematic branches in the data you're testing on, because the function returns early?

`uprn` I think is safe because we have the `if uprn else ""`, but I think in general adding defaults across the board is sensible.

help_to_heat/portal/download_views.py
```diff
@@ -279,9 +279,9 @@ def add_extra_row_data(referral, exclude_pii=False):
     created_at = referral.created_at.astimezone(london_tz)
 
     if not exclude_pii:
-        contact_number = row.get("contact_number")
+        contact_number = row.get("contact_number", "")
         contact_number = '="' + contact_number + '"'
-        uprn = row.get("uprn")
+        uprn = row.get("uprn", "")
```
jdgage, line 284: Should probably be consistent here with the if/else and defaults. Don't have strong preferences where we go within that.

jdgage (changes requested):

help_to_heat/portal/download_views.py
samyou-softwire: for python this shouldn't error, since tuples aren't constrained to a single type so it is possible a tuple of strings contains None. in general when we designed calculate_eligibility it was done resiliently since we were keeping open the idea of calling it on an incomplete form

help_to_heat/portal/download_views.py
```diff
@@ -281,7 +281,7 @@ def add_extra_row_data(referral, exclude_pii=False):
     if not exclude_pii:
         contact_number = row.get("contact_number", "")
         contact_number = '="' + contact_number + '"'
-        uprn = row.get("uprn", "")
+        uprn = row.get("uprn")
         uprn = '="' + str(uprn) + '"' if uprn else ""
```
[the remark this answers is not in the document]
samyou-softwire: to avoid errors I needed to wrap uprn in a str(), and did the same for contact_number too just in case

jdgage:
Last thing is the PR to remove the endpoint, if you could please make that

help_to_heat/portal/download_views.py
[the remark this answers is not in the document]
jdgage: Oh yes, I'd missed that - thanks!

jdgage (approved):

PR #291, opened 2024-03-21
PC-912 Remove admin URL to extract all referrals

samyou-softwire:
pre-empting its removal from the project


jdgage (changes requested):
Let's get rid of the function in `download_views` as well, as there will be nothing pointing at it (avoids someone getting confused at some point)

samyou-softwire:
> @samyou-softwire I have a memory that we agreed to keep the endpoint, and I documented that somewhere. Do you remember where? Would like to close this if I've remembered it right.

it's in the comments of [PC-1171](https://beisdigital.atlassian.net/browse/PC-1171)

PR #295, opened 2024-03-28
PC 898 partial fix specific backlinks

Glenn-Clarke:
Added specific backlink cases to make the backlink on the following pages in the following scenarios route the user back to the /address page

From the benefits page when the following two conditions are fulfilled:
- The user selects that they have a Park Home
- There is no EPC found for their Park Home, or they've selected that they live in Scotland

From the property type page when the following two conditions are fulfilled:
- The user selects that they live in Social Housing
- There is no EPC found for their home, or they've selected that they live in Scotland

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -644,7 +644,6 @@ class PropertyTypeView(PageView):
             return super().get_prev_next_urls(session_id, page_name)
 
 
-
 @register_page("property-subtype")
 class PropertySubtypeView(PageView):
     def get_context(self, request, session_id, *args, **kwargs):
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -634,7 +634,6 @@ class PropertyTypeView(PageView):
         session_data = interface.api.session.get_session(session_id)
         own_property = session_data.get("own_property")
         country = session_data.get("country")
-        epc = session_data.get("epc")
         epc_rating = session_data.get("epc_rating", "Not found")
 
         if own_property == "No, I am a social housing tenant" and (country == "Scotland" or epc_rating == "Not found"):
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -595,6 +595,19 @@ class BenefitsView(PageView):
             return redirect("frontdoor:page", session_id=session_id, page_name="property-type")
         return super().handle_post(request, session_id, page_name, data, is_change_page)
 
+    def get_prev_next_urls(self, session_id, page_name):
+        session_data = interface.api.session.get_session(session_id)
+        park_home = session_data.get("park_home")
+        country = session_data.get("country")
+        epc_rating = session_data.get("epc_rating", "Not found")
+
+        if park_home == "Yes" and (country == "Scotland" or epc_rating == "Not found"):
+            _, next_page_url = get_prev_next_urls(session_id, page_name)
+            prev_page_url = reverse("frontdoor:page", kwargs=dict(session_id=session_id, page_name="address"))
+            return prev_page_url, next_page_url
+        else:
+            return super().get_prev_next_urls(session_id, page_name)
+
 
 @register_page("household-income")
 class HouseholdIncomeView(PageView):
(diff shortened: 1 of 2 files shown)


samyou-softwire (changes requested):

help_to_heat/frontdoor/views.py
```diff
@@ -595,6 +595,19 @@ def handle_post(self, request, session_id, page_name, data, is_change_page):
             return redirect("frontdoor:page", session_id=session_id, page_name="property-type")
         return super().handle_post(request, session_id, page_name, data, is_change_page)
 
+    def get_prev_next_urls(self, session_id, page_name):
+        session_data = interface.api.session.get_session(session_id)
+        park_home = session_data.get("park_home")
+        country = session_data.get("country")
+        epc_rating = session_data.get("epc_rating", "Not found")
+
+        if park_home == "Yes" and (country == "Scotland" or epc_rating == "Not found"):
```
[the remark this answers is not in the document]
Glenn-Clarke: This is a good point, I'll make this change and check it still works as intended.

help_to_heat/frontdoor/views.py
```diff
@@ -595,6 +595,19 @@ def handle_post(self, request, session_id, page_name, data, is_change_page):
...
+            _, next_page_url = get_prev_next_urls(session_id, page_name)
```
jdgage, line 605: Is there a subtle difference between this and `super().get_prev_next_urls` below?

help_to_heat/frontdoor/views.py
```diff
@@ -617,6 +630,19 @@ class PropertyTypeView(PageView):
     def get_context(self, *args, **kwargs):
         return {"property_type_options": schemas.property_type_options_map}
 
+    def get_prev_next_urls(self, session_id, page_name):
+        session_data = interface.api.session.get_session(session_id)
+        own_property = session_data.get("own_property")
+        country = session_data.get("country")
+        epc_rating = session_data.get("epc_rating", "Not found")
+
+        if own_property == "No, I am a social housing tenant" and (country == "Scotland" or epc_rating == "Not found"):
```
jdgage, line 639: Sanity-checking both changes as the ticket is ambiguous I think: is this definitely an `and`, not an `or`?

Glenn-Clarke: image
I haven't checked personally, but I believe it's "and" going by the backlink flow chart from the investigation.

help_to_heat/frontdoor/views.py
Glenn-Clarke: I don't believe there's a functional difference at present, as super().get_prev_next_urls immediately passes the inputs through to get_prev_next_urls, and returns the result of that function without any changes.

I'm reusing the style established for special cases within the method on line 709.

samyou-softwire (approved):

tests/test_frontdoor.py
```diff
@@ -302,6 +302,7 @@ def test_back_button():
     assert form["country"] == "England"
 
 
+
```
jdgage, line 305: Guess this was unintentional? But are the tests missing somewhere here?

jdgage (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: Yes, fixing now. Changelist issues

tests/test_frontdoor.py
```diff
@@ -302,6 +302,224 @@ def test_back_button():
     assert form["country"] == "England"
 
 
+@unittest.mock.patch("help_to_heat.frontdoor.interface.EPCApi", MockNotFoundEPCApi)
+@unittest.mock.patch("help_to_heat.frontdoor.interface.OSApi", MockOSApi)
+@utils.mock_os_api
+def test_benefits_back_button_with_park_home_and_no_epc():
```
jdgage, line 308: Nitpick: could we have these test names refer to what the app should do? Easier to reason with when making changes in the future that way

tests/test_frontdoor.py
```diff
@@ -302,6 +302,224 @@ def test_back_button():
...
+
+    page = page.click(contains="Back")
+
+    assert page.has_one('h1:contains("What is the property\'s address?")')
+
+
+@unittest.mock.patch("help_to_heat.frontdoor.interface.OSApi", MockOSApi)
+@utils.mock_os_api
+def test_property_type_back_button_with_social_housing_and_scotland():
+    client = utils.get_client()
```
jdgage, line 475: Only if it ends up being really quick, otherwise we can leave it for next time:

There's a fair bit of common steps between these tests. One of the things I'm keen to do with this file is extract duplicated logic to helper methods like `fill_in_such_and_such_a_page/pages`. I'm hoping we can chip away at is from time to time as we go. Would it be easy to do something like that here?

jdgage (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: Are you envisioning chunky helpers with several questions answered, or smaller ones like:

`def country_selection(page: interface.api.session, country: str) -> interface.api.session:
    form = page.get_form()
    form["country"] = country
    return form.submit().follow()
`

jdgage: As discussed, something more like the last thing. If you've already got something in that direction then do throw it up here and I'll have a look, but I think this is good to merge as-is now.

jdgage (approved):

tests/test_frontdoor.py
Glenn-Clarke: The bit I've put together isn't working initially, so might be best to just merge as is now. 
But I've got some ideas now for if we come back to this.

PR #297, opened 2024-04-02
PC-900 Fix broken backlinks for Loft insulation

samyou-softwire:
closes [PC-900](https://beisdigital.atlassian.net/browse/PC-900)

adds a back link exception in a similar style to [PC-898](https://github.com/UKGovernmentBEIS/help-to-heat-GBIS/pull/295), depending on the answer to the loft question

adds a verifying test, though it is quite large since it needs to nearly (but not quite) complete the whole flow. another option would be to utilise a flow finishing function & have the test press the back button twice

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -737,6 +737,18 @@ class SummaryView(PageView):
             return supplier_redirect
         return super().handle_post(request, session_id, page_name, data, is_change_page)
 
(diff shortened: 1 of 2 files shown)


jdgage (approved):
Think this is fine. The test is pretty wordy, yes, but that's a bigger problem that hopefully we can address soon. It would be a bit more manageable (and readable) if we had some 'complete screen'/'complete flow' common methods.

PR #299, opened 2024-04-04
PC-911 Add incorrect password or username error message to portal login

samyou-softwire:
closes [PC-911](https://beisdigital.atlassian.net/browse/PC-911)

adds new error messages as discussed with client

folds custom error messages into the already built error function. as far as I can tell the signature of error is not important so adding a new parameter should be fine

in order to show links to the user for the email, the safe tag is applied when showing messages. since the messages do not interpolate user input this should be fine

image

image

tests are included


help_to_heat/portal/authentication_views.py
```diff
@@ -32,10 +32,21 @@ def _make_hash_value(self, user, timestamp):
 @require_http_methods(["GET", "POST"])
 class CustomLoginView(MethodDispatcher):
     template_name = "account/login.html"
-    error_message = "Something has gone wrong.  Please contact your team leader."
+    credentials_error_message = (
+        "Login failed - please check your credentials. If you believe they are correct, "
+        "contact your team leader or our support team at "
+        '<a href="mailto:eligibilitycheckersupport-cai@energysecurity.gov.uk'
+        '">eligibilitycheckersupport-cai@energysecurity.gov.uk</a>'
```
jdgage, line 39:
```suggestion
        '">eligibilitycheckersupport-cai@energysecurity.gov.uk</a>.'
```

help_to_heat/portal/authentication_views.py
```diff
@@ -32,10 +32,21 @@ def _make_hash_value(self, user, timestamp):
...
+    )
+    invite_link_error_message = (
+        "Login failed - please accept the invitation you were sent via email. If you have "
+        "not received this, contact your team leader or our support team at <a "
+        'href="mailto:eligibilitycheckersupport-cai@energysecurity.gov.uk'
+        '">eligibilitycheckersupport-cai@energysecurity.gov.uk</a>'
```
jdgage, line 45:
```suggestion
        '">eligibilitycheckersupport-cai@energysecurity.gov.uk</a>.'
```

help_to_heat/templates/allauth/account/base.html
```diff
@@ -85,7 +85,7 @@
             <div>
               <ul class="govuk-list">
                 {% for message in messages %}
-                  <li>{{message}}</li>
+                  <li>{{message | safe}}</li>
```
jdgage, line 88: Are we definitely always in control of the message we're setting here?

help_to_heat/portal/authentication_views.py
```diff
@@ -32,10 +32,21 @@ def _make_hash_value(self, user, timestamp):
 @require_http_methods(["GET", "POST"])
 class CustomLoginView(MethodDispatcher):
     template_name = "account/login.html"
-    error_message = "Something has gone wrong.  Please contact your team leader."
+    credentials_error_message = (
```
jdgage, line 35: Bit puzzled by this: why are these tuples, and why have you split the strings where you've split them?

jdgage (changes requested):

help_to_heat/portal/authentication_views.py
samyou-softwire: this was an IDE suggestion to stop any single line being too long, since theyre not comma seperated theyre not evalutated as tuples

help_to_heat/templates/allauth/account/base.html
samyou-softwire: So to confirm, the message is only set from inside authentication_views.py and is set to the following values (line numbers included)
- Please enter an email and password. (59)
- Login failed - please check your credentials. If you believe they are correct... (69)
- Login failed - please accept the invitation you were sent via email. If you have... (65)
- Something has gone wrong.  Please contact your team leader. (74) (110) (165) (214) (226) (230)
- Please enter an email. (125)
- Passwords must match. (170) (298)
- **ValidationError output messages** (175) (310)
- Please fix the errors and try again (177)
- Password successfully set. (182)
- Please enter the otp (220)
- This link is not valid. It may have expired or have already been used. Please try again. (275) (280) (302)

the messages that derive from ValidationError are the only particularly scary ones, though from doing some inspection they do not provide user input directly. examples include:
- This password is too common.
- This password is entirely numeric.
- This password is too short. It must contain at least %(min_length)d character.

though if it's not worth the risk the email can be left as plain text

jdgage: I think I'd feel more comfortable if we left it as plain text. It's vanishingly unlikely, but this would leave us open to a supply chain attack where one of our dependencies (django or marshmallow by the looks of things) gets updated in such a way that we end up injecting some unsafe HTML. And of course the open source supply chain has been in the news this week with `xz`! Would need to be more carefully considered to do properly, and I don't think that's in scope for this ticket.

jdgage (changes requested):

jdgage (approved):

PR #302, opened 2024-04-10
PC-899: Fix backlinks for Special supplier warning pages (Shell, Bulb and Utility Warehouse)

Glenn-Clarke:
Added special cases to own-property backlinks where supplier is Bulb, Utility Warehouse, or Shell to go back to their appropriate warning page.

Added tests for these special cases too, to ensure they're working correctly.

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -345,11 +345,15 @@ class OwnPropertyView(PageView):
             return prev_page_url, next_page_url
         elif request_supplier == "Utility Warehouse":
             _, next_page_url = get_prev_next_urls(session_id, page_name)
-            prev_page_url = reverse("frontdoor:page", kwargs=dict(session_id=session_id, page_name="utility-warehouse-warning-page"))
+            prev_page_url = reverse(
+                "frontdoor:page", kwargs=dict(session_id=session_id, page_name="utility-warehouse-warning-page")
+            )
             return prev_page_url, next_page_url
         elif request_supplier == "Shell":
             _, next_page_url = get_prev_next_urls(session_id, page_name)
-            prev_page_url = reverse("frontdoor:page", kwargs=dict(session_id=session_id, page_name="shell-warning-page"))
+            prev_page_url = reverse(
+                "frontdoor:page", kwargs=dict(session_id=session_id, page_name="shell-warning-page")
+            )
             return prev_page_url, next_page_url
         else:
             return super().get_prev_next_urls(session_id, page_name)
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -340,15 +340,15 @@ class OwnPropertyView(PageView):
         session_data = interface.api.session.get_session(session_id)
         request_supplier = session_data.get("supplier")
         prev_page_url, next_page_url = super().get_prev_next_urls(session_id, page_name)
-        if request_supplier == "Bulb, now part of Octopus Energy":
-            prev_page_url = reverse("frontdoor:page", kwargs=dict(session_id=session_id, page_name="bulb-warning-page"))
-        elif request_supplier == "Utility Warehouse":
+        acquired_supplier_warning_pages = {
+            "Bulb, now part of Octopus Energy": "bulb-warning-page",
+            "Utility Warehouse": "utility-warehouse-warning-page",
+            "Shell": "shell-warning-page",
+        }
+        if request_supplier in acquired_supplier_warning_pages:
             prev_page_url = reverse(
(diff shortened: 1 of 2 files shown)


tests/test_frontdoor.py
```diff
@@ -304,6 +304,43 @@ def test_back_button():
...
+    assert page.status_code == 302
+    page = page.follow()
+
+    assert page.status_code == 200
+    session_id = page.path.split("/")[1]
+    assert uuid.UUID(session_id)
+
+    form = page.get_form()
+    form["country"] = "England"
+    page = form.submit().follow()
```
samyou-softwire, line 327: seems a good time to use `_check_page`? ie
```suggestion
    page = _check_page(page, "country", "country", "England")
```

samyou-softwire (changes requested):

help_to_heat/frontdoor/views.py
```diff
@@ -336,6 +336,28 @@ def handle_post(self, request, session_id, page_name, data, is_change_page):
...
+            )
+            return prev_page_url, next_page_url
+        elif request_supplier == "Shell":
+            _, next_page_url = get_prev_next_urls(session_id, page_name)
+            prev_page_url = reverse(
+                "frontdoor:page", kwargs=dict(session_id=session_id, page_name="shell-warning-page")
+            )
+            return prev_page_url, next_page_url
+        else:
+            return super().get_prev_next_urls(session_id, page_name)
```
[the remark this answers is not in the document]
Glenn-Clarke: I like this solution, I think it was this sort of simplification I was trying to find yesterday, so I'm glad you've spotted it.

tests/test_frontdoor.py
Glenn-Clarke: Also checked the supplier lines below to check page

help_to_heat/frontdoor/views.py
```diff
@@ -336,6 +336,22 @@ def handle_post(self, request, session_id, page_name, data, is_change_page):
 
         return redirect("frontdoor:page", session_id=session_id, page_name=next_page_name)
 
+    def get_prev_next_urls(self, session_id, page_name):
+        session_data = interface.api.session.get_session(session_id)
+        request_supplier = session_data.get("supplier")
+        prev_page_url, next_page_url = super().get_prev_next_urls(session_id, page_name)
+        if request_supplier == "Bulb, now part of Octopus Energy":
```
jdgage, line 343: I think I'd go one further than Sammy here, and have a dictionary of `acquired_supplier_warning_pages` say:
```
acquired_supplier_warning_pages = {
    "Bulb, now part of Octopus Energy": "bulb-warning-page",
    ...
}
```

Then we'd only need one
```
if request_supplier in acquired_supplier_warning_pages:
    prev_page_url = reverse(
                "frontdoor:page", kwargs=dict(session_id=session_id, page_name=acquired_supplier_warning_pages[request_supplier])
    )
```
rather than a long chain if `elif`s (that may get longer, who knows).

tests/test_frontdoor.py
```diff
@@ -304,6 +304,41 @@ def test_back_button():
...
 
+@pytest.mark.parametrize(
+    ("supplier_name", "expected_text"),
+    [
+        ("Shell", "Shell is now owned by Octopus Energy."),
+        ("Bulb, now part of Octopus Energy", "Bulb is now owned by Octopus Energy."),
+        ("Utility Warehouse", "Referral requests from UW customers will be managed by E.ON Next"),
+    ],
+)
+def test_own_property_back_button_with_supplier_should_return_to_supplier_warning_page(supplier_name, expected_text):
```
jdgage, line 315: Are we testing the happy path somewhere else?

jdgage (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: What is the happy path? @jdgage

Glenn-Clarke: I've looked into it, and it seems test_back_button might fit the bill?
```
    form = page.get_form()
    assert form["own_property"] == "Yes, I own my property and live in it"

    form = page.get_form()
    assert form["supplier"] == "Utilita"

    form = page.get_form()
    assert form["country"] == "England"
```

jdgage: Happy path in this case is that the back button does what we expect when there's not a warning page. So yes, it looks like that test covers it.

In general, it's context dependent, but it's usually going to be the journey the user takes through a system when everything's 'normal', there aren't errors, warnings, etc. Most of the time it will be the most common path through the system, but that's not necessarily true.

jdgage (approved):
Nice!

PR #303, opened 2024-04-12
PC-971 Park homes questions incorrectly served when user changes answers from home owner to social housing

samyou-softwire:
closes [PC-971](https://beisdigital.atlassian.net/browse/PC-971)

the main issue is that the park home check took priority over the social housing check for deciding which flow the user is on. swapping the check fixes the issue of incorrect flow

also to fix the issue of the question appearing on the check answers page a new exception has been added to hide it when needed

added some tests to verify this. contains a few upgrades for tests
- functions to interface with the change buttons
- add a park home option when answering house questions

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -161,9 +161,9 @@ def get_prev_next_page_name(page_name, session_id=None):
         is_social_housing = session_data.get("own_property") == "No, I am a social housing tenant"
         receives_benefits = session_data.get("benefits") == "Yes"
 
+    # This question is asked first, so this path should take precedence in case users have
+    # gone back and changed their answers
     if is_social_housing:
-        # This question is asked first, so this path should take precedence in case users have
-        # gone back and changed their answers
         mapping = schemas.page_prev_next_map_social_housing
         order = schemas.page_order_social_housing
     elif is_park_home:
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -161,7 +161,7 @@ def get_prev_next_page_name(page_name, session_id=None):
         is_social_housing = session_data.get("own_property") == "No, I am a social housing tenant"
         receives_benefits = session_data.get("benefits") == "Yes"
 
-    if is_social_housing:
+    if is_social_housing:  # This question is asked first, so this path should be checked first to take precedence
         mapping = schemas.page_prev_next_map_social_housing
         order = schemas.page_order_social_housing
     elif is_park_home:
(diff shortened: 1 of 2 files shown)


help_to_heat/frontdoor/views.py
```diff
@@ -735,9 +735,12 @@ def show_question(self, session_data, question):
         question_answered = question in session_data and question in schemas.summary_map
         if not question_answered:
             return False
-        if question == "property_type" or question == "property_subtype":
+        if question in ["property_type", "property_subtype"]:
             property_type = self.get_answer(session_data, "property_type")
             return property_type != "Park home"
+        if question in ["park_home", "park_home_main_residence"]:
+            own_property = self.get_answer(session_data, "own_property")
+            return own_property != "No, I am a social housing tenant"
```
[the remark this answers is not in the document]
Glenn-Clarke: image
According to the flows I've seen, the question for park homes is asked on every path except social housing.

help_to_heat/frontdoor/views.py
```diff
@@ -161,12 +161,12 @@ def get_prev_next_page_name(page_name, session_id=None):
         is_social_housing = session_data.get("own_property") == "No, I am a social housing tenant"
         receives_benefits = session_data.get("benefits") == "Yes"
 
-    if is_park_home:
-        mapping = schemas.page_prev_next_map_park_home
-        order = schemas.page_order_park_home
-    elif is_social_housing:
+    if is_social_housing:  # This question is asked first, so this path should be checked first to take precedence
```
jdgage, line 164:
```suggestion
    if is_social_housing:  # This question is asked first, so this path should take precedence in case users have gone back and changed their answers
```

tests/test_frontdoor.py
```diff
@@ -1374,6 +1374,7 @@ def test_on_check_page_back_button_goes_to_correct_location(has_loft_insulation)
     else:
         assert page.has_one("h1:contains('Do you have a loft that has not been converted into a room?')")
 
+
```
jdgage, line 1377: Looks like a stray newline here

jdgage (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: The linter wants this here for check-code, should I still remove this?

help_to_heat/frontdoor/views.py
jdgage: Yes, I think that would be clearer

tests/test_frontdoor.py
jdgage: Oh I see, it was wrong before was it - no problem then

PR #307, opened 2024-04-24
PC 994 prevent incompatible loft answers in change page

Glenn-Clarke:
Change to prevent users changing (or even seeing) follow up questions regarding their loft, should they specify that they do not have one.

When the user answers yes: 
image
When the user answers no: 
image

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -765,7 +765,7 @@ class SummaryView(PageView):
             return epc_rating != "Not found" and accept_suggested_epc == "Yes"
         if question in ["loft_access", "loft_insulation"]:
             loft_answer = self.get_answer(session_data, "loft")
-            return loft_answer != "No, I do not have a loft or my loft has been converted into a room"
+            return loft_answer == "Yes, I have a loft that has not been converted into a room"
         else:
             return True
 
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -763,6 +763,9 @@ class SummaryView(PageView):
             epc_rating = session_data.get("epc_rating", "Not found")
             accept_suggested_epc = session_data.get("accept_suggested_epc")
             return epc_rating != "Not found" and accept_suggested_epc == "Yes"
+        if question in ["loft_access", "loft_insulation"]:
+            loft_answer = self.get_answer(session_data, "loft")
+            return loft_answer != "No, I do not have a loft or my loft has been converted into a room"
         else:
             return True
 
--- a/tests/test_frontdoor.py
+++ b/tests/test_frontdoor.py
@@ -1463,7 +1463,7 @@ def test_switching_path_to_social_housing_does_not_ask_park_home_questions_again
 
 @unittest.mock.patch("help_to_heat.frontdoor.interface.EPCApi", MockEPCApi)
 def test_on_check_answers_page_changing_to_social_housing_hides_park_home_question():
-    _check_page, page, session_id = setup_client_and_page()
+    _check_page, page, session_id = _setup_client_and_page()
 
     # Answer main flow
     page = _answer_house_questions(page, session_id, benefits_answer="Yes")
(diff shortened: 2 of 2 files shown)


help_to_heat/frontdoor/views.py
```diff
@@ -763,6 +763,9 @@ def show_question(self, session_data, question):
             epc_rating = session_data.get("epc_rating", "Not found")
             accept_suggested_epc = session_data.get("accept_suggested_epc")
             return epc_rating != "Not found" and accept_suggested_epc == "Yes"
+        if question in ["loft_access", "loft_insulation"]:
+            loft_answer = self.get_answer(session_data, "loft")
+            return loft_answer != "No, I do not have a loft or my loft has been converted into a room"
```
samyou-softwire, line 768: seems a bit safer (and a bit more expressive) in this binary case to write as show these questions if loft answer is yes

tests/test_frontdoor.py
```diff
@@ -280,6 +285,12 @@ def _assert_change_button_is_hidden(page, page_name):
     assert change is None
 
 
+def _assert_change_button_is_not_hidden(page, page_name):
+    elements = page.all("a:contains('Change')")
+    change = next(filter(lambda el: f"/{page_name}/change/" in el.attrib["href"], elements), None)
+    assert change is not None
```
samyou-softwire, line 291: considering this is now the 3rd time this logic is expressed may be time to extract to a helper

tests/test_frontdoor.py
```diff
@@ -1473,3 +1483,35 @@ def test_on_check_answers_page_changing_to_social_housing_hides_park_home_questi
...
+
+    # Answer main flow
+    page = _answer_house_questions(page, session_id, benefits_answer="Yes", has_loft=True)
+
+    # press back to get to summary page
+    page = page.click(contains="Back")
+    assert page.has_one("h1:contains('Check your answers')")
+
+    _assert_change_button_is_not_hidden(page, "loft-access")
+    _assert_change_button_is_not_hidden(page, "loft-insulation")
```
samyou-softwire, line 1508: this is a good check - probably worth adding this to the previous `test_on_check_answers_page_changing_to_social_housing_hides_park_home_question` too

samyou-softwire (changes requested):

samyou-softwire (approved):
LGTM

tests/test_frontdoor.py
```diff
@@ -1465,6 +1481,13 @@ def test_on_check_answers_page_changing_to_social_housing_hides_park_home_questi
     page = page.click(contains="Back")
     assert page.has_one("h1:contains('Check your answers')")
 
+    if park_home:
+        _assert_change_button_is_not_hidden(page, "park-home")
+        _assert_change_button_is_not_hidden(page, "park-home-main-residence")
+    else:
+        _assert_change_button_is_not_hidden(page, "park-home")
+        _assert_change_button_is_hidden(page, "park-home-main-residence")
```
jdgage, line 1489: We could pull the first assertion outside the if/else, since it's the same in both places.

I kind of think the `park-home-main-residence` assertion doesn't belong here: it's sort of orthogonal to what the test is testing. It's really testing the logic around hiding that bit depending on the park home answer, which has its own test (I think? Should do!). Unit tests should really only test one thing, and shouldn't overlap: that means if a test fails then you know immediately what's gone wrong, and don't have to spend time debugging the test itself

(I'd maybe even drop the `park_home` parameter from the test now I look at it, and just cover one case arbitrarily)

jdgage (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: > (I think? Should do!).

I've split the test into two to ensure we're not losing any coverage :)

tests/test_frontdoor.py
```diff
@@ -1473,3 +1482,57 @@ def test_on_check_answers_page_changing_to_social_housing_hides_park_home_questi
...
+    assert page.has_text("Do you have a loft that has not been converted into a room?")
+
+    page = _check_page(page, "loft", "loft", "No, I do not have a loft or my loft has been converted into a room")
+    assert page.has_one("h1:contains('Check your answers')")
+
+    _assert_change_button_is_hidden(page, "loft-access")
+    _assert_change_button_is_hidden(page, "loft-insulation")
+
+
+def setup_client_and_page():
```
jdgage, line 1529:
```suggestion
def _setup_client_and_page():
```
By convention since this is a 'private' method.

jdgage (approved):

PR #313, opened 2024-05-14
PC-1053: migrate to poetry package management

Glenn-Clarke:
Migrated to use Poetry as a dependency manager.
- Dependency groups have been used to separate dependencies used on deployed versions and dev
- Removed update dependencies makefile
- Moved some lines in dockerfiles to optimise caching
- Updated documentation to help initial setup.


.github/workflows/portal-checks.yml
```diff
@@ -34,7 +34,9 @@ jobs:
     - name: Install Dependencies
       run: |
         python3 -m pip install --upgrade pip setuptools
-        pip install -r requirements-dev.lock
+        pip install poetry
```
jdgage, line 37: Can this happen with the previous line as well? Don't know enough to say myself

docker/tests/Dockerfile
```diff
@@ -25,12 +20,20 @@ RUN chmod +x /start.sh
...
 
-COPY ./ /app
-
 COPY ./envs/web.template ./envs/web
 
+COPY poetry.lock pyproject.toml /app/
+
 WORKDIR /app/
 
+# not recommended in docs to install poetry globally but as this is a VM this is fine
```
jdgage, line 29: This comment doesn't look right - GBIS doesn't run in a VM

docker/tests/Dockerfile
```diff
@@ -25,12 +20,20 @@ RUN chmod +x /start.sh
...
+RUN pip install poetry && \
+  # no need for venv in dev container
```
jdgage, line 31: Likewise: no dev container (also contradicts the readme changes?)

docker/web/Dockerfile
```diff
@@ -32,10 +25,18 @@ RUN chmod +x /entrypoint.sh
 
 COPY ./envs/web.template ./envs/web
 
-COPY . /app
+COPY poetry.lock pyproject.toml /app/
 
 WORKDIR /app/
 
+# not recommended in docs to install poetry globally but as this is a VM this is fine
```
jdgage, line 32: Same

docker/web/Dockerfile
```diff
@@ -32,10 +25,18 @@ RUN chmod +x /entrypoint.sh
...
 COPY ./envs/web.template ./envs/web
 
-COPY . /app
+COPY poetry.lock pyproject.toml /app/
 
 WORKDIR /app/
 
+# not recommended in docs to install poetry globally but as this is a VM this is fine
+RUN pip install poetry && \
+  # no need for venv in dev container
```
jdgage, line 34: Same

pyproject.toml
```diff
@@ -1,3 +1,58 @@
+[tool.poetry]
+package-mode = false
+[tool.poetry.group.dev.dependencies]
```
jdgage, line 3: How did you settle on these groups? I'm not sure what we're doing now is equivalent to what we had before

jdgage (changes requested):

pyproject.toml
Glenn-Clarke: I settled on using three groups:
- Dev  - dependencies exclusive to requirements-dev.lock, explicitly mentioned in the docs
- Deployment - dependencies exclusive to requirements.lock
- No group (main group) - dependencies shared between the two

The intent is that main group loads in both scenarios, while deployment is only included when explicitly specified during `poetry install --with deployment` and dev seemed to have special functionality within Poetry according to the docs, so its only included in dev environments.

Have I misunderstood how this works?

docker/tests/Dockerfile
Glenn-Clarke: Yes, sorry, I think this was an artefact from the framework we were using for this ticket. I'll remove these

.github/workflows/portal-checks.yml
Glenn-Clarke: Yes, I think so.

README.md
```diff
@@ -99,6 +101,7 @@ the string is used (rather than once on server load).
 
 ### Context
 
+In some cases the English string alone is not enough to identify the translation (e.g. "Close" as in quit and "Close" as
```
samyou-softwire, line 104: this line is duplicated

docker/web/Dockerfile
```diff
@@ -32,10 +25,16 @@ RUN chmod +x /entrypoint.sh
...
-COPY . /app
+COPY poetry.lock pyproject.toml /app/
 
 WORKDIR /app/
 
+RUN pip install poetry && \
+  poetry config virtualenvs.create false && \
+  poetry install --with deployment
+
+COPY ./ /app
```
samyou-softwire, line 36: probably worth keeping consistent on whether we use a trailing / or not throughout the files (the previous `COPY . /app` does not)

samyou-softwire (changes requested):

.github/workflows/portal-checks.yml
jdgage: Oh interesting, did we not need to install pip at all? Already there?

README.md
```diff
@@ -3,9 +3,11 @@
 Formerly known as "Help to Heat".
 
 ## Initial Setup
-
-1. From the project root, run `cp envs/web.template envs/web`
-2. Populate the OS API key into the OS_API_KEY variable in `envs/web`. The key can be found in Keeper. Please only use a single value (i.e. `["my_key_here"]`), rather than all of the keys that are stored in Keeper, in order to avoid needing to rotate all of the keys if your local environment is accidentally leaked.
+1. [Install Poetry](https://python-poetry.org/docs/) on your machine
+2. [Install Python 3.9.13](https://www.python.org/downloads/release/python-3913/)
+3. Set up a virtual environment within the project for Poetry to manage your dependencies. A guide for Pycharm can be found [here](https://www.jetbrains.com/help/pycharm/poetry.html).
```
jdgage, line 8: Oh looks like my comment went missing 😢 meant to write here before

I don't think we need this given everything's running in a container? I'd hope not anyway - I don't particularly want to require people to install an old version of Python.

docker/web/Dockerfile
jdgage: You need the trailing slash when there's multiple sources specified (e.g. line 28), so needs to be that one we stick with

pyproject.toml
jdgage: Got you - I think it was that missing `--with dev` that was the issue as you already spotted.

docker/tests/Dockerfile
jdgage: I do think the comments might still be warranted if you can reword

jdgage (changes requested):

README.md
Glenn-Clarke: Is there a reason to not do our dev work on the same version for compatibility, so our code runs how we expect it to?

Additionally, to use poetry, we need a compatible version of python installed with respect to the toml. Therefore to work with poetry on newer python versions, we'd have to loosen the requirement in the toml instead of locking it down as much as I've done currently.

jdgage: Yes, we do want to do our dev work on the same version of everything, but that's what Docker gives us. It should be the case that when you run the site/tests/whatever locally you're using Docker to do that - is that not true in practice?

Glenn-Clarke: That's true, yes, while we're running GBIS locally, as the dockerfile has the version of python to be used in the image at the top of each file (which currently means it runs on 3.9.17). 
Does that mean we're good to loosen up the restriction in the .toml to something like "~3.9", so we can have our local version of python be more recent? Because our local version needs to be compatible with the restriction we've declared in the .toml to actually use Poetry. 

I can't find in the docs if poetry would complain that we've installed a dependency that is incompatible with 3.9, but works with 3.10 for example, and it wouldn't become clear until it tries to run in docker? Would it complain that the new dependency might not work with the potential minimum python version?

jdgage: I don't think you need to have a local version of python at all: it's in the container you spin up and everything runs in there.

I don't know if we even need to have Python in `pyproject.toml` at all? I'd kind of expect it to be implicit...

Glenn-Clarke: We need a local version of python so our IDE can know what packages are installed.

To be clear, this doesn't affect running tests or the site. Its more a question of us needing it to perform dev work in the IDE prior to running anything. This local version we're using to perform dev work must be compatible with the toml file's specified range.

If I remove the python requirement entirely from the toml, I end up with issues surrounding the resolution of package versions. 
image
I've consulted with Sammy, and we think we have to specify it in the toml to resolve this.

Glenn-Clarke: I've made an update now to note in the poetry section's header that it is optional :)

README.md
```diff
@@ -1,11 +1,15 @@
...
+### (Optional) Poetry Setup 
+If you want to be able to make changes to the dependency list using Poetry, or install the dependencies locally (so you can click into them in your IDE, or use autocomplete) perform these steps:
+1. [Install Poetry](https://python-poetry.org/docs/) on your machine
+2. [Install Python 3.9.13](https://www.python.org/downloads/release/python-3913/)
+3. Set up a virtual environment within the project for Poetry to manage your dependencies. A guide for Pycharm can be found [here](https://www.jetbrains.com/help/pycharm/poetry.html).
+### Main steps
+4. From the project root, run `cp envs/web.template envs/web`
+5. Populate the OS API key into the OS_API_KEY variable in `envs/web`. The key can be found in Keeper. Please only use a single value (i.e. `["my_key_here"]`), rather than all of the keys that are stored in Keeper, in order to avoid needing to rotate all of the keys if your local environment is accidentally leaked.
```
jdgage, line 12: Tiny thing now: can these steps move up under Initial Setup where they were before and the optional section be below them separately? I think it's easier to miss them this way round.

jdgage (changes requested):

README.md
Glenn-Clarke: That's true, I've swapped them around now :)

jdgage (approved):

PR #323, opened 2024-07-05
PC-1135: add error causing endpoint

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1135)

# Description

ticket to add an error causing endpoint.

**this ticket must be followed up by merging #324 before release, this is endpoint should not ever be available on staging or prod**


jdgage (commented):
Could we (also) have something that's a bit more organic? e.g. set up an unhandled null reference exception, maybe nested in a couple of functions?

jdgage (approved):

PR #333, opened 2024-07-23
PC-1218: Fix key error if referral is missing PII

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1218)

# Description

add a check before popping key

this resolves a KeyError if the key is not in the object, which can happen for invalid referrals

this does not fix the root issue, but does make the export handler more resilient

# Checklist

### DIFF
--- a/help_to_heat/portal/download_views.py
+++ b/help_to_heat/portal/download_views.py
@@ -271,7 +271,8 @@ def add_extra_row_data(referral, exclude_pii=False):
 
     if exclude_pii:
         for key in pii_keys:
-            row.pop(key)
+            if key in row.keys():
+                row.pop(key)
 
     eligibility = calculate_eligibility(row)
     epc_date = row.get("epc_date")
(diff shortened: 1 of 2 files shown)


help_to_heat/portal/download_views.py
```diff
@@ -271,7 +271,8 @@ def add_extra_row_data(referral, exclude_pii=False):
 
     if exclude_pii:
         for key in pii_keys:
-            row.pop(key)
+            if key in row.keys():
+                row.pop(key)
```
jdgage, line 275: Sanity-checking: if not `key in row.keys()` then there's no data there and we'll just end up with a blank cell anyway?

samyou-softwire: yeah, there's the following line in the xlsx writer:
```py
    for row_num, referral_data in enumerate(rows):
        for col_num, entry in enumerate(column_headings):
            to_write = referral_data.get(entry) or ""
            worksheet.write(row_num + 1, col_num, to_write)
```
so should default to "" if not present (it's not uncommon for columns to be missing, loft access often is)

jdgage (approved):

PR #344, opened 2024-08-22
PC-1229: Single source of truth function for previous page

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1229)

# Description

implements single source of truth function for previous page

the function itself is quite small, as it uses the existing forwards logic

testing for the function however needs to be quite thorough, each test needs to construct an entire valid questionnaire up to that point for the back links to work

# Checklist


help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -0,0 +1,30 @@
+from help_to_heat.frontdoor.consts import (
+    country_page,
+    govuk_start_page,
+    unknown_page,
+)
+from help_to_heat.frontdoor.routing.forwards_routing import get_next_page
+
+start_page = country_page
+max_route_length = 100
```
jdgage, line 9: This is presumably arbitrary to catch loops? Can we mark it as such? How much higher is this than the actual realistic maximum?

help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -0,0 +1,30 @@
...
+    while len(route) < max_route_length:
+        if route_page == current_page:
+            return route[-1]
+
+        if route_page == unknown_page:
+            return unknown_page
+
+        next_page, redirect = get_next_page(route_page, answers)
+
+        if not redirect:
```
jdgage, line 25: Think this warrants explanation from your comment on the other PR.

help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -0,0 +1,30 @@
...
+)
+from help_to_heat.frontdoor.routing.forwards_routing import get_next_page
+
+start_page = country_page
+max_route_length = 100
+
+
+def get_prev_page(current_page, answers):
+    route_page = start_page
+    route = [govuk_start_page]
```
jdgage, line 14: Probably doesn't make a real difference with the quantities of data we're talking about, but I believe a deque would be more performant here.

tests/routing/__init__.py
jdgage, line 1: I think at this point it would be good to have a comment explaining the purpose of the generators in here and how they should be maintained.

jdgage (changes requested):

help_to_heat/frontdoor/routing/backwards_routing.py
samyou-softwire: sure, at a quick count the realistic maximum is around 30

help_to_heat/frontdoor/routing/backwards_routing.py
samyou-softwire: sure

help_to_heat/frontdoor/routing/backwards_routing.py
samyou-softwire: have changed to a deque

tests/routing/__init__.py
samyou-softwire: have written some explanation

will add some more general thoughts on routing practices on swiki when [PC-1228](https://beisdigital.atlassian.net/browse/PC-1228) is completed

jdgage (approved):

PR #348, opened 2024-09-02
PC-1231: Get route function

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1231)

# Description

implements get route function

`get_prev_page` can now be implemented in terms of `get_route`

govuk_start_page is now more properly treated as a page, with handles for next & prev page

# Checklist


help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -14,20 +15,13 @@
...
+        return unknown_page
 
-        route_page = next_page
+    try:
+        route = get_route(answers, current_page)
 
-    return unknown_page
+        return route[-2]
+    except CouldNotCalculateRouteException:
+        return unknown_page
```
jdgage, line 27: Does this get automatically logged?

help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -14,20 +15,13 @@
...
+    # there is no previous page to this
+    if current_page == govuk_start_page:
+        return unknown_page
 
-        route_page = next_page
+    try:
+        route = get_route(answers, current_page)
 
-    return unknown_page
+        return route[-2]
```
jdgage, line 25: Warrants a comment to save someone having to work it out themselves

help_to_heat/frontdoor/routing/__init__.py
```diff
@@ -0,0 +1,34 @@
...
+
+from help_to_heat.frontdoor.consts import govuk_start_page, unknown_page
+from help_to_heat.frontdoor.routing.forwards_routing import get_next_page
+
+# in case of infinite loop ensure a route can't go on forever
+# currently the longest route is under 30 pages
+max_route_length = 100
+
+
+def get_route(answers, target_page, start_page=None):
```
jdgage, line 11: Think this could do with a comment explaining what it's producing and probably a rename: route being a loaded term (I'd expect a function with this name to return a relative path by default). Should `target` be `current` here, and `current_page` renamed to something else below? `target makes me think I'm trying to go somewhere.

help_to_heat/frontdoor/routing/forwards_routing.py
```diff
@@ -215,6 +219,10 @@ def get_next_page(current_page, answers):
     return _unknown_response
 
 
+def _govuk_start_page_next_page(answers):
+    return country_page
```
jdgage, line 223:
```suggestion
def _govuk_start_page_next_page():
    return country_page
```

jdgage (changes requested):

help_to_heat/frontdoor/routing/backwards_routing.py
samyou-softwire: no - if I remove the catch it would be

jdgage: Worth logging something in that case then: sort of thing we'd want to catch if it's happening

help_to_heat/frontdoor/routing/forwards_routing.py
jdgage: Yes I think so

help_to_heat/frontdoor/routing/__init__.py
```diff
@@ -0,0 +1,53 @@
...
+
+from help_to_heat.frontdoor.consts import govuk_start_page, unknown_page
+from help_to_heat.frontdoor.routing.forwards_routing import get_next_page
+
+# in case of infinite loop ensure a route can't go on forever
+# currently the longest route is under 30 pages
+max_route_length = 100
+
+
+def calculate_route(answers, to_page, from_page=None):
```
jdgage, line 11: Would still like to find something other than "route" here, if possible. Or at least adding some more words to clarify. It's clear what it's going when you read the comment here, but I don't think it is in the other file where this gets called.

samyou-softwire: talked with Glenn a bit, perhaps journey would be a better term?

help_to_heat/frontdoor/routing/backwards_routing.py
samyou-softwire: probably easiest to remove the try entirely - the error would be logged by our global error handling

help_to_heat/frontdoor/routing/forwards_routing.py
samyou-softwire: done

help_to_heat/frontdoor/routing/backwards_routing.py
```diff
@@ -14,20 +15,15 @@
...
+        return unknown_page
+
+    try:
+        journey = calculate_journey(answers, current_page)
+
+        # if the page has been found, the last item in the list (index -1) will be the current page
+        # so, the item before the last one (index -2) will be the previous page
+        return journey[-2]
+    except CouldNotCalculateJourneyException:
+        return unknown_page
```
jdgage, line 29: I'm confused where we are with this: thought either this was being removed to just throw and bubble up, or we were logging here?

In which case is there any helpful message we should include with the error, for diagnostic purposes?

samyou-softwire: yeah I'm a little undecided here on returning a telling value that there was an error vs. throwing. the reason I've gone this way is
- `get_next_page` doesn't throw an error if it can't find the next page so it feels more natural to follow suit
- there are times where we don't want this to throw. on the manual address pages (back again!) as we're hard coding a link to them, when they're loaded the back links can't be automatically generated. I added some code to the view to override this but if this function were to throw then the page would error before the override logic runs.
if we do want to preserve the error throwing behaviour I could add some exceptions to the routing logic to ensure in the case of the error manual pages it doesn't throw

samyou-softwire: I agree, I've made it so this throws and the places where this can be ignored are in the schema, so in the cases where we know the back url the function isn't called

jdgage (approved):

PR #355, opened 2024-09-12
PC-1295: Set up connection for open epc api

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1295)

# Description

Added the API Key and endpoint to the secrets template

Adding to feature branch as this would break dev without the following tickets.

# Checklist


envs/web.template
jdgage, line 1: There's some other places these variables get used in actually allowing the code to use the API, which will be what the pipeline is unhappy about.

jdgage (changes requested):

envs/web.template
Glenn-Clarke: I think I've got them all, from some further searching 😅

help_to_heat/settings.py
```diff
@@ -253,9 +253,8 @@ def show_toolbar(request):
 }
 
 OS_API_KEY = env.str("OS_API_KEY")
-EPC_API_BASE_URL = env.str("EPC_API_BASE_URL")
-EPC_API_CLIENT_ID = env.str("EPC_API_CLIENT_ID")
-EPC_API_CLIENT_SECRET = env.str("EPC_API_CLIENT_SECRET")
```
jdgage, line 258: Were these being used somewhere in the app? I'd expect them to be consumed somewhere we're actually using the API

Glenn-Clarke: Ah I wasn't sure that was part of the scope. I've made some further changes now.
I'm going to sort the tests, but could you give the code a skim? (understand time is tight today)

jdgage (approved):

help_to_heat/settings.py
jdgage: Might not have been, but I suppose the auth part is useful to have, and it's only going to save future effort for the other bits. If it turns out to be a load of work or you're getting too close to actually doing the implementation then I suppose you could leave the current implementation as-is and have the new stuff commented out to be used on the implementation tickets?

PR #366, opened 2024-09-30
PC-1313: Better messaging for routing errors

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1313)

# Description

adds a new page for an error calculating the users' journey

the case that throws this error is now checked explicitly (rather than relying on back button routing error)

the exceptions to this rule have been named more appropriately

translations have been added

# Checklist

# Screenshots

image
image


Glenn-Clarke (approved):
LGTM

help_to_heat/frontdoor/views.py
```diff
@@ -371,6 +375,13 @@ def get(self, request, session_id, page_name, unsaved_data=None, errors=None, is
         if data_with_get_answers != data:
             save_answer(session_id, page_name, data_with_get_answers)
 
+        if page_name not in schemas.routing_overrides:
+            try:
+                # ensure the user has answers to complete the journey to this page
+                calculate_journey(answers, to_page=page_name)
+            except CouldNotCalculateJourneyException:
+                return redirect("/sorry-journey")
```
jdgage, line 383: Do we want to log a warning or something else below an error here now? I forget whether that's covered by another ticket.

samyou-softwire: good point; we do get the choice here for whether we want to or not

I'd err on doing so for now (I'll update the code), then we'll monitor going forwards whether we still want to or not (covered by that other ticket)

help_to_heat/frontdoor/views.py
```diff
@@ -371,6 +375,14 @@ def get(self, request, session_id, page_name, unsaved_data=None, errors=None, is
         if data_with_get_answers != data:
             save_answer(session_id, page_name, data_with_get_answers)
 
+        if page_name not in schemas.routing_overrides:
+            try:
+                # ensure the user has answers to complete the journey to this page
+                calculate_journey(answers, to_page=page_name)
+            except CouldNotCalculateJourneyException as e:
+                logger.exception(e)
```
jdgage, line 383: Ok, so to confirm, this is the equivalent to what we currently do, we just send them to a different page?

samyou-softwire: yes, to be specific:
- currently, if you try to view a page that's not on your route, you will see an error page
- this happens when trying to calculate the previous page
- calculating the previous page requires finding a route to your page, and going 1 back
- the code will fail on trying to find a route to your page
- this has been moved to an explicit check
- so functionality should be identical to before

jdgage: Yes, and in particular we'll still get an error logged for it?

jdgage (approved):

help_to_heat/frontdoor/views.py
samyou-softwire: yes (the one in this diff)

PR #367, opened 2024-10-02
PC-1298: Add new fields to expose to energy suppliers

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1298)

# Description

adds code to extract the new epc fields and display them in exported EPCs

add a new API call for the epc recommendations

updates the two sets of csv headers (pii & no pii) to have appropriate values

generates the improvements text

removes the now unnecessary `property_main_heat_source` data field, as this information can be extracted from the epc details. in general sets the precedent to query epc info from there

# Checklist

# Screenshots

image


help_to_heat/frontdoor/views.py
```diff
@@ -637,14 +638,13 @@ def save_post_data(self, data, session_id, page_name):
 
         uprn = epc_details.get("uprn")
         address = epc_details.get("address")
-        heat_source = epc_details.get("mainheat-description")
 
         epc_data = {
             lmk_field: lmk,
             address_field: address,
             epc_details_field: epc_details,
+            recommendations_field: recommendations,
```
samyou-softwire, line 646: save all recommendations rather than say a formatted string, in case we want to change how we show these to energy suppliers in the future

help_to_heat/frontdoor/views.py
```diff
@@ -637,14 +638,13 @@ def save_post_data(self, data, session_id, page_name):
...
             uprn_field: uprn if uprn is not None else "",
-            property_main_heat_source_field: heat_source if heat_source is not None else "",
```
samyou-softwire, line 647: chose to deprecate this as it's now non-standard and may be confusing why this field only is exported like this

help_to_heat/portal/download_views.py
```diff
@@ -75,11 +135,58 @@
     "loft",
     "loft_access",
     "loft_insulation",
-    "Property main heat source",
```
samyou-softwire, line 78: this appeared to never be set to a value

Glenn-Clarke (approved):

help_to_heat/frontdoor/epc_api.py
```diff
@@ -8,21 +8,28 @@
 
 
 class EPCApi:
+    def __init__(self):
+        self.base_url = settings.OPEN_EPC_API_BASE_URL
```
jdgage, line 12: Can we just use `settings.OPEN_EPC_API_BASE_URL` directly?

help_to_heat/frontdoor/interface.py
```diff
@@ -544,6 +544,10 @@ def get_epc_details(self, lmk):
         epc_api = EPCApi()
         return epc_api.get_epc_details(lmk)
 
+    def get_epc_recommendations(self, lmk):
+        epc_api = EPCApi()
+        return epc_api.get_epc_recommendations(lmk)
```
jdgage, line 549: Feels like this would be the appropriate level to combine all the data we get back from the API (unless there's a reason not to?) - would merge this with the above method and return an object with both `epc_details` and `recommendations`.

help_to_heat/frontdoor/views.py
jdgage: Is this something we should note on the ticket/check with DESNZ?

help_to_heat/portal/download_views.py
```diff
@@ -297,7 +405,89 @@ def add_extra_row_data(referral, exclude_pii=False):
...
         "referral_id": referral.formatted_referral_id,
+        "user_selected_property_type": user_selected_property_type,
+    }
+
+    epc_data = row.get(epc_details_field) or {}
+
+    # main heat source used to come from a specific field in the user data
+    # now it is sent via the epc details object
+    # use the legacy field if available, else draw from EPC
+    property_main_heat_source = row.get(property_main_heat_source_field) or epc_data.get("mainheat-description")
```
jdgage, line 416: For what it's worth I think it equally raises questions we we now treat it differently here (not suggesting it's wrong necessarily, just that it's going to stick out either way)

help_to_heat/portal/download_views.py
```diff
@@ -297,7 +405,89 @@ def add_extra_row_data(referral, exclude_pii=False):
...
     }
+
+    if not exclude_pii:
+        row = {
+            **row,
+            "lmk_key": epc_data.get("lmk-key"),
+            "address_line_1": epc_data.get("address1"),
+            "address_line_2": epc_data.get("address2"),
+            "address_line_3": epc_data.get("address3"),
+        }
```
jdgage, line 479: We should have a record on the ticket (agreed with DESNZ) of what fields we're excluding. Can you open up that discussion please?

jdgage (changes requested):

help_to_heat/portal/download_views.py
samyou-softwire: can do

help_to_heat/frontdoor/interface.py
samyou-softwire: have merged them

help_to_heat/portal/download_views.py
samyou-softwire: I agree, though I think it's more sensible to keep the higher traffic routing code saner and handle it here

does the comment explain this well enough?

help_to_heat/frontdoor/epc_api.py
samyou-softwire: sure

help_to_heat/portal/download_views.py
jdgage: Yes think so

jdgage (approved):
Happy with this once PII is agreed

samyou-softwire:
> Happy with this once PII is agreed

PII has been confirmed by DESNZ

PR #370, opened 2024-10-10
PC-1348: Add date field to EPC

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1348)

# Description

Added a "date of issue" section to the epc page

# Checklist

# Screenshots

image

image


help_to_heat/frontdoor/views.py
```diff
@@ -807,10 +808,18 @@ def build_extra_context(self, request, session_id, page_name, data, is_change_pa
         epc_band = epc.get("current-energy-rating")
         epc_date = epc.get("lodgement-date")
 
+        try:
+            working_epc_date = datetime.strptime(epc_date, "%Y-%m-%d")
+            month_name = month_names[working_epc_date.month - 1]
+            gds_epc_date = f"{working_epc_date.strftime('%-d')} {month_name} {working_epc_date.strftime('%Y')}"
+        except ValueError:
```
jdgage, line 815: So we started discussing this before - remind me why we'd get this?

help_to_heat/templates/frontdoor/epc.html
```diff
@@ -46,6 +46,12 @@ <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title
         <p class="govuk-body">{{ epc_rating.upper() }}</p>
       {% endif %}
       
+      <p class="govuk-body govuk-!-font-weight-bold govuk-!-margin-bottom-1">{{_("Date of issue")}}</p>
+
+      {% if gds_epc_date %}
```
jdgage, line 51: Do we always expect to have this?

Glenn-Clarke: Yes, we should be able to always expect this. I've removed this check now

help_to_heat/frontdoor/views.py
jdgage: Well, it's a fair point. How confident are we that what we get here will always be in the format we expect. Would it be worth having the date fall back to being blank if (for whatever reason) there was an error parsing it, rather than the whole page erroring and blocking the journey?

Glenn-Clarke: - For Scotland, I think it should *always* be in the format we expect, given the typing enforced on the date in our database will ensure this for Scotland. 
- For England, I think we actually might just get the date from the Open API as a string, looking again. In this case, the date's format is enforced by their Database presumably (though it *could* be a string in their database). Perhaps we ought to include the check then?

Previously I defaulted to just displaying the EPC date in the format it appears from the database (since this would be what the energy suppliers see, too). But it would mean it wouldn't be GDS compliant in those cases. Is being blank better, since it's GDS compliant?

Glenn-Clarke: That sounds good to me. I've expanded the epc.html if block to contain the "Date of Issue" title too.

jdgage (approved):

PR #371, opened 2024-10-14
PC-1346: Change "I can't find my address..." link to radio button

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1346)

# Description

- adds new functionality to the radios macro to allow for a fallback 'or' value to be shown
- when selected, set the appropriate routing answer and continue as normal
- previous overrides to make the URL work can be removed
- update error message
- add new tests for the two possible values of the fallback & fix existing ones
- miro board updating is pending access

# Checklist

- [ ] If I have made any changes to website flow, I have updated the [Flow Miro Board](https://miro.com/app/board/uXjVK4AgQ3E=/)

# Screenshots

image
image
image


help_to_heat/frontdoor/views.py
```diff
@@ -628,14 +628,27 @@ def build_extra_context(self, request, session_id, page_name, data, is_change_pa
             }
             for a in address_and_rrn_details
         )
+
+        fallback_option = {
+            "value": "enter-manually",
```
jdgage, line 633: Can we pull this magic string out to a constant on the class?

help_to_heat/frontdoor/views.py
```diff
@@ -691,18 +704,32 @@ def build_extra_context(self, request, session_id, page_name, data, is_change_pa
             }
             for a in addresses
         )
+
+        fallback_option = {
+            "value": "enter-manually",
```
jdgage, line 709: Same here

jdgage (changes requested):
Have you got a screenshot of when there's no addresses found?

help_to_heat/frontdoor/consts.py
```diff
@@ -169,10 +169,12 @@
 duplicate_uprn_field = "uprn_is_duplicate"
 epc_found_field = "epc_found"
 lmk_field = "lmk"
+lmk_field_enter_manually = "enter-manually"
```
jdgage, line 172: Maybe not worth moving them now, but given they're only used within the class in question I don't think they need to be pulled out to this level (and having one global 'all constants' file is kind of messy)

jdgage (commented):
I feel like we shouldn't have a radio button if there are no addresses found? Seems quite user-hostile to make them check a radio button _and_ press a button to go down the only route available to them. Where did that requirement come from? I think we should question it.

help_to_heat/frontdoor/consts.py
samyou-softwire: perhaps the constants file could be reviewed (the constants are a bit inconsistently spread between multiple files); could raise a backlog ticket to look into it

in this case since this is a form answer it will be passed through to the routing functions and so could feasibly be imported & reasoned with there, so I think this should be hoisted

samyou-softwire:
> I feel like we shouldn't have a radio button if there are no addresses found? Seems quite user-hostile to make them check a radio button and press a button to go down the only route available to them. Where did that requirement come from? I think we should question it.

this isn't a specific requirement. perhaps for now I could make this work as it did before in this case and we can raise a backlog ticket to think further about how best to present this to the user

image

jdgage:
> > I feel like we shouldn't have a radio button if there are no addresses found? Seems quite user-hostile to make them check a radio button and press a button to go down the only route available to them. Where did that requirement come from? I think we should question it.
> 
> this isn't a specific requirement. perhaps for now I could make this work as it did before in this case and we can raise a backlog ticket to think further about how best to present this to the user
I think yes let's do that, although I don't know if it needs reviewing. As long as it takes one click I think it's fine.

samyou-softwire:
> > > I feel like we shouldn't have a radio button if there are no addresses found? Seems quite user-hostile to make them check a radio button and press a button to go down the only route available to them. Where did that requirement come from? I think we should question it.
> > 
> > 
> > this isn't a specific requirement. perhaps for now I could make this work as it did before in this case and we can raise a backlog ticket to think further about how best to present this to the user
> I think yes let's do that, although I don't know if it needs reviewing. As long as it takes one click I think it's fine.
for what its worth the prior flow wasn't great (the user can see the continue button but pressing it doesn't work), so I do think it could do with a touch up

samyou-softwire:
have raised as [PC-1378](https://beisdigital.atlassian.net/browse/PC-1378)

help_to_heat/frontdoor/consts.py
samyou-softwire: have raised as [PC-1379](https://beisdigital.atlassian.net/browse/PC-1379)

jdgage (approved):

Glenn-Clarke:
Fixed some merge issues (since my ticket also edited the translations)

PR #373, opened 2024-10-16
PC-1373: Migration for new columns in Scottish EPC table

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1373)

# Description

adds new columns to scottish epc table

regardless of their format, all new fields are strings. there is never a requirement to treat them as numbers or dates

field lengths are based on viewing live data and are intentionally over-cautious

in order to check that details can be imported the epc importing script has been updated. to unify with GBIS it is now included in this repository. the resultant csv was able to be imported into my local database using pycharm's csv importing tool

# Checklist


scripts/scottish-epc-formatter/README.md
jdgage, line 1: I think this wants a line or two describing what it's for

help_to_heat/portal/models.py
```diff
@@ -165,3 +165,53 @@ class ScottishEpcRating(utils.TimeStampedModel):
...
+    roof_energy_efficiency = models.CharField(max_length=16, null=True)
+    roof_environmental_efficiency = models.CharField(max_length=16, null=True)
+    lighting_description = models.CharField(max_length=64, null=True)
+    lighting_energy_efficiency = models.CharField(max_length=16, null=True)
+    lighting_environmental_efficiency = models.CharField(max_length=16, null=True)
+    mechanical_ventilation = models.CharField(max_length=64, null=True)
+    construction_age_band = models.CharField(max_length=16, null=True)
+    tenure = models.CharField(max_length=64, null=True)
+    improvements = models.TextField(max_length=1024, null=True)
+    alternative_improvements = models.TextField(max_length=1024, null=True)
```
jdgage, line 217: Is this long enough?

Glenn-Clarke: @jdgage 
I'm not sure what the maximum length of these fields can be, since I think they do something similar to us in that this is a collection of all the improvement strings.
I don't think we _need_ to assign one since we're using Postgresql: 
https://docs.djangoproject.com/en/5.1/ref/models/fields/#django.db.models.CharField.max_length
I'm not sure how VarChars are dealt with though, would this take up more memory than limited length fields?

jdgage: Would they: yes. Does it matter: probably not a great deal. Generally it's a good idea to put an upper limit on column size, but only if you can be confident about it (and I'm not sure we can in this case?). It matters less if you're not querying against that column, which is the case here.

Glenn-Clarke:
@samyou-softwire I've pushed a change to remove the character limits, but I've been unable to do the migration as I'm having issues with Django on my system. I think this just needs to have a new migration created.

help_to_heat/portal/models.py
```diff
@@ -165,3 +165,53 @@ class ScottishEpcRating(utils.TimeStampedModel):
     uprn = models.CharField(max_length=12, primary_key=True)
     rating = models.CharField(max_length=32, choices=epc_rating_choices)
     date = models.DateField(blank=True, null=True)
+    property_type = models.CharField(max_length=32, null=True)
+    address1 = models.CharField(max_length=64, null=True)
+    address2 = models.CharField(max_length=64, null=True)
+    address3 = models.CharField(max_length=64, null=True)
```
jdgage, line 171: I would probably go longer with these too

Glenn-Clarke: I think 512 should be sufficient.

I've found some discussion online that the longest address in the royal mail PAF service is about 370 characters. (But I think that might encompass all three lines.)

jdgage (approved):

PR #378, opened 2024-10-23
PC-1375: Remove text length limits from Scottish EPC text

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1375)

# Description

we decided that the text lengths for some potentially variable length EPC fields weren't worth it, so they are removed in this PR

made them django TextFields to allow for this/ the documentation says that a CharField can have unlimited length but in practice django would not allow it

# Checklist


jdgage (commented):
I think for obvious cases, e.g. postcode, EPC rating(?) we should still have max length

samyou-softwire:
> I think for obvious cases, e.g. postcode, EPC rating(?) we should still have max length

I suppose without further research into the data I'd be wary, we have little control over the data format. e.g. for EPC rating I'm fairly certain an observed option is 'None' or wtte. we could overestimate those but it feels like it defeats the point

jdgage:
I'd be fine with having things be either 'short text' i.e. 20 char limit or 'long text' i.e. no limit. I think we do need to be mindful of how much disk space this table is going to take up.

samyou-softwire:
> I'd be fine with having things be either 'short text' i.e. 20 char limit or 'long text' i.e. no limit. I think we do need to be mindful of how much disk space this table is going to take up.

have pared back the changes a bit now to mostly text fields

jdgage (approved):

PR #383, opened 2024-10-28
PC-1374: Link new Scottish EPC database columns to Supplier CSV

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1374)

# Description

exposes the data in scottish epc to the csv

does this by saving an object similar to an E&W csv to the session

to think about; is this fine or should we have a more stringent schema? I wasn't able to add one easily as
- I'd like to keep the format consistent across E&W & scotland
- the E&W layout cannot easily be made into a marshmallow schema as many keys have hyphens. hyphens cannot be a field name in python
- we could make a new schema, but then the E&W data would need to change format & the csv exporter would need more cases to handle

# Checklist

### DIFF
--- a/help_to_heat/frontdoor/interface.py
+++ b/help_to_heat/frontdoor/interface.py
@@ -525,14 +525,74 @@ class Address(Entity):
 
 
 class EPC(Entity):
-    @with_schema(load=GetScottishEPCSchema, dump=EPCSchema)
+    @with_schema(load=GetScottishEPCSchema)
     def get_epc_scotland(self, uprn):
         try:
             epc = portal.models.ScottishEpcRating.objects.get(uprn=uprn)
         except portal.models.ScottishEpcRating.DoesNotExist:
             return {}
 
-        return {"uprn": epc.uprn, "rating": epc.rating, "date": epc.date}
+        improvements = epc.improvements
+        alternative_improvements = epc.alternative_improvements
+        all_improvements = improvements + (
+            "|| Alternatives: " + alternative_improvements if alternative_improvements else ""
+        )
+
+        return {
+            # follows column names in help_to_heat/frontdoor/mock_epc_api_data/sample_epc_response.json
+            "uprn": epc.uprn,
+            "current-energy-rating": epc.rating,
+            "lodgement-date": epc.date,
+            "property-type": epc.property_type,
+            "address1": epc.address1,
+            "address2": epc.address2,
+            "address3": epc.address3,
+            "postcode": epc.postcode,
+            "building-reference-number": epc.building_reference_number,
+            "potential-energy-rating": epc.potential_rating,
+            "current-energy-efficiency": epc.current_energy_efficiency_rating,
+            "potential-energy-efficiency": epc.potential_energy_efficiency_rating,
+            "built-form": epc.built_form,
+            "inspection-date": epc.inspection_date,
+            "local-authority-label": epc.local_authority,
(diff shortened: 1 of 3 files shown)


help_to_heat/frontdoor/interface.py
```diff
@@ -525,14 +525,74 @@ def parse_dpa_to_address(self, dpa):
...
             epc = portal.models.ScottishEpcRating.objects.get(uprn=uprn)
         except portal.models.ScottishEpcRating.DoesNotExist:
             return {}
 
-        return {"uprn": epc.uprn, "rating": epc.rating, "date": epc.date}
+        improvements = epc.improvements or ""
+        alternative_improvements = epc.alternative_improvements or ""
+        all_improvements = improvements + (
+            "Alternatives: " + alternative_improvements if alternative_improvements else ""
+        )
```
jdgage, line 539: Do you need both fallbacks for the alternatives?

help_to_heat/frontdoor/views.py
```diff
@@ -746,8 +746,7 @@ def save_post_data(self, data, session_id, page_name):
         if country == country_field_scotland and uprn is not None:
             epc = interface.api.epc.get_epc_scotland(uprn)
             if epc != {}:
-                epc_details = {"current-energy-rating": epc.get("rating"), "lodgement-date": epc.get("date")}
-                data[epc_details_field] = epc_details
```
jdgage, line 750: Would prefer to just pass the fields we care about to the template, while we're changing this

jdgage (changes requested):

help_to_heat/frontdoor/views.py
samyou-softwire: for E&W? in the scotland case here all the fields passed here are ones we care about

jdgage: Ah sorry, I had this backwards - never mind

help_to_heat/frontdoor/interface.py
```diff
@@ -525,14 +525,74 @@ def parse_dpa_to_address(self, dpa):
...
             epc = portal.models.ScottishEpcRating.objects.get(uprn=uprn)
         except portal.models.ScottishEpcRating.DoesNotExist:
             return {}
 
-        return {"uprn": epc.uprn, "rating": epc.rating, "date": epc.date}
+        improvements = epc.improvements
+        alternative_improvements = epc.alternative_improvements
+        all_improvements = improvements + (
+            "Alternatives: " + alternative_improvements if alternative_improvements else ""
+        )
```
jdgage, line 539: Two questions:
1. Should there be a space and/or some punctuation before 'Alternatives'?
2. Is `epc.improvements` ever going to be `None`? We'll need a fallback if so

samyou-softwire: It's not something I'm particularly concerned for, the load from the CSV means anything empty should become an empty string. If we want to account for it, then I'd probably want to `or ""` the other fields too

jdgage: I think throwing something in before 'Alternatives' would still be a good idea, so we don't have to make assumptions about the content of `epc.improvements`

samyou-softwire: Perhaps `|| Alternatives `? Like a separator but with more weight to it

jdgage (approved):

PR #390, opened 2024-10-31
PC-1415: Fix pipelines not failing on code format fail

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1415)

# Description

this ensures `black` and `isort` emit a non-zero exit code on formatting fail
`flake8` does this automatically so no changes needed here

job pre these changes: https://github.com/UKGovernmentBEIS/help-to-heat-GBIS/actions/runs/11611247770/job/32332179284
job post these changes: https://github.com/UKGovernmentBEIS/help-to-heat-GBIS/actions/runs/11611170314/job/32331954594

# Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,10 @@ add-suppliers:
 
 # -------------------------------------- Code Style  -------------------------------------
 
+.PHONY: format-python-code
+format-python-code:
+	docker compose -f tests/docker-compose.yml build format-code-help-to-heat && docker compose -f tests/docker-compose.yml run --no-deps --rm format-code-help-to-heat
+
 .PHONY: check-python-code
 check-python-code:
 	docker compose -f tests/docker-compose.yml build check-code-help-to-heat && docker compose -f tests/docker-compose.yml run --no-deps --rm check-code-help-to-heat
--- a/docker/tests/check-code.sh
+++ b/docker/tests/check-code.sh
@@ -1,8 +1,13 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
-set -o nounset
 
-black .
-isort .
+if [[ $1 = "--format" ]]; then
+    black .
+    isort .
+else
+    black --check --diff .
+    isort --check --diff .
+fi
+
 flake8 .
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -18,6 +18,10 @@ add-suppliers:
 check-python-code:
 	docker compose -f tests/docker-compose.yml build check-code-help-to-heat && docker compose -f tests/docker-compose.yml run --no-deps --rm check-code-help-to-heat
 
+.PHONY: check-python-code-ci
+check-python-code-ci:
+	docker compose -f tests/docker-compose.yml build check-code-help-to-heat && docker compose -f tests/docker-compose.yml run --no-deps --rm check-code-ci-help-to-heat
+
```
jdgage, line 24: Think we can do some renaming across the board here: e.g. `/check-code.sh --check` reads a bit weirdly, and without the option we're fixing/formatting rather than checking (this was a problem before). So would favour something like changing all the old `check`s to `format` and then just `check-code` where you've got `check-code-ci` now.

jdgage (changes requested):

docker/tests/check-code.sh
```diff
@@ -1,8 +1,13 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
-set -o nounset
```
samyou-softwire, line 4: don't think this is needed & it causes issues if $1 is unset (flag isn't passed)

Makefile
samyou-softwire: have done some reshuffling, think this makes a bit more sense. have updated the readme as the task to run locally has changed

tests/docker-compose.yml
```diff
@@ -20,12 +20,17 @@ services:
...
+    env_file:
+      - ../envs/tests
+    entrypoint: bash /check-code.sh
+    volumes:
+      - ../:/app/:z
+
+  check-code-ci-help-to-heat:
+    extends:
+      service: check-code-help-to-heat
+    entrypoint: bash /check-code.sh --check
```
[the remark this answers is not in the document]
samyou-softwire: sure

jdgage (approved):

PR #402, opened 2024-11-25
PC-1465: stop old epcs showing in epc select

Glenn-Clarke:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1465)

# Description

Filters the returned list of EPCs for E&W to only include the most recent EPC for each UPRN present in the list.

# Checklist


tests/test_frontdoor.py
```diff
@@ -2112,6 +2112,43 @@ def test_epc_page_shows_epc_info():
...
+
+    form = page.get_form()
+    form["building_name_or_number"] = "22"
+    form["postcode"] = "FL23 4JA"
+    page = form.submit().follow()
+
+    data = interface.api.session.get_answer(session_id, page_name="address")
+    assert len(data["address_and_lmk_details"]) == 2
+    assert data["address_and_lmk_details"][0]["lmk-key"] != "3333333333333333333333333333333333"
+    assert data["address_and_lmk_details"][1]["lmk-key"] != "3333333333333333333333333333333333"
```
samyou-softwire, line 2149: perhaps better to use the defined consts for the keys here (& maybe page names too?)

help_to_heat/utils.py
```diff
@@ -249,3 +250,14 @@ def get_current_and_next_month_names(month_names):
...
 def get_current_scottish_epc_cutoff_and_next_dump_month_names(month_names):
     return month_names[9], month_names[1]
+
+
+def get_most_recent_epc_per_uprn(address_and_lmk_details):
+    most_recent_address_and_lmk_details = []
+    address_and_lmk_details.sort(key=lambda x: x["uprn"], reverse=True)
+
+    for _, group in itertools.groupby(address_and_lmk_details, lambda x: x["uprn"]):
+        item = max(list(group), key=lambda x: datetime.strptime(x["lodgement-date"], "%Y-%m-%d"))
```
samyou-softwire, line 260: does this work fine for epcs with no uprn?

samyou-softwire, line 260: is it necessary to wrap `group` in a list here? would assume its already iterable

samyou-softwire, line 260: could the variable name be improved here?

help_to_heat/utils.py
```diff
@@ -249,3 +250,14 @@ def get_current_and_next_month_names(month_names):
 # this is to tell the user the first month we have no EPCs for, and the next month we expect to perform a dump
 def get_current_scottish_epc_cutoff_and_next_dump_month_names(month_names):
     return month_names[9], month_names[1]
+
+
+def get_most_recent_epc_per_uprn(address_and_lmk_details):
+    most_recent_address_and_lmk_details = []
+    address_and_lmk_details.sort(key=lambda x: x["uprn"], reverse=True)
```
samyou-softwire, line 257: may be better to use [`sorted`](https://www.w3schools.com/python/ref_func_sorted.asp) here to avoid mutating the list we pass in?

tests/test_frontdoor.py
```diff
@@ -2112,6 +2112,43 @@ def test_epc_page_shows_epc_info():
...
+
+    assert page.has_text("Do you live in a park home")
+    page = _check_page(page, "park-home", "park_home", "No")
+
+    form = page.get_form()
+    form["building_name_or_number"] = "22"
+    form["postcode"] = "FL23 4JA"
+    page = form.submit().follow()
+
+    data = interface.api.session.get_answer(session_id, page_name="address")
```
samyou-softwire, line 2146: for what its worth it may be more in the spirit of these tests to check what text appears on the page rather than the underlying answers. maybe both?

samyou-softwire (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: I think you're right that the page's appearing text should be checked, but the underlying data is critical to determine whether it's the correct one of the two identical addresses showing.

help_to_heat/utils.py
Glenn-Clarke: I didn't realise this was an issue, I'll add a case where this ignores no UPRN cases, since as discussed it hasn't been seen yet in practice, but is a niche possibility.

help_to_heat/utils.py
samyou-softwire, line 257: seems like this lambda could be replaced with an [`attrgetter`](https://docs.python.org/3/howto/sorting.html#operator-module-functions-and-partial-function-evaluation)

Glenn-Clarke: Oh nice! I hadn't seen this before, I'll use it below as well

help_to_heat/utils.py
samyou-softwire: for what its worth I've added this caveat to our docs here: https://softwiretech.atlassian.net/wiki/spaces/Support/pages/20562509848/DESNZ+GBIS+Help+to+Heat+EPCs#EPC-UPRNs

help_to_heat/utils.py
```diff
@@ -249,3 +250,17 @@ def get_current_and_next_month_names(month_names):
 # this is to tell the user the first month we have no EPCs for, and the next month we expect to perform a dump
 def get_current_scottish_epc_cutoff_and_next_dump_month_names(month_names):
     return month_names[9], month_names[1]
+
+
+def get_most_recent_epc_per_uprn(address_and_lmk_details):
+    most_recent_address_and_lmk_details = []
+    sorted_address_and_lmk_details = sorted(address_and_lmk_details, key=lambda x: x["uprn"], reverse=True)
+
+    for uprn, group in itertools.groupby(sorted_address_and_lmk_details, lambda x: x["uprn"]):
```
samyou-softwire, line 259: I suppose it's not clear at this time whether a missing UPRN will come through as an empty string or be vacant entirely; probably safer to use the `.get()` method rather than indexing to provide a sensible default & avoid an error being thrown (probably `""` to align with the below line)

probably safer to use `.get()` in all cases here

tests/test_frontdoor.py
```diff
@@ -2112,6 +2116,46 @@ def test_epc_page_shows_epc_info():
...
+    page = _check_page(page, "park-home", "park_home", "No")
+
+    form = page.get_form()
+    form["building_name_or_number"] = "22"
+    form["postcode"] = "FL23 4JA"
+    page = form.submit().follow()
+
+    data = interface.api.session.get_answer(session_id, page_name=address_page)
+
+    assert page.has_one("label:contains('22 Acacia Avenue, Upper Wellgood, Fulchester, FL23 4JA')")
```
samyou-softwire, line 2152: we would expect two addresses to be present? feels like this is what the below line indicates

samyou-softwire (changes requested):

tests/test_frontdoor.py
Glenn-Clarke: We expect only one instance of this visible address value. 

The line below is checking the data to ensure that neither is the excluded one on an LMK level, since we can't tell which of the above address line is shown at the current screen.

Was trying to keep the test focused on this address, given the focus of this test is about pinpointing that the newest EPC is found specifically.

samyou-softwire (approved):

help_to_heat/utils.py
```diff
@@ -249,3 +250,17 @@ def get_current_and_next_month_names(month_names):
...
 def get_current_scottish_epc_cutoff_and_next_dump_month_names(month_names):
     return month_names[9], month_names[1]
+
+
+def get_most_recent_epc_per_uprn(address_and_lmk_details):
+    most_recent_address_and_lmk_details = []
+    sorted_address_and_lmk_details = sorted(address_and_lmk_details, key=lambda x: x.get("uprn", ""), reverse=True)
+
+    for uprn, group in itertools.groupby(sorted_address_and_lmk_details, lambda x: x.get("uprn", "")):
+        if uprn == "":
```
jdgage, line 260: Think we should document here why we do this (i.e. what's in our docs) and our justification for including them all in this case.

tests/test_frontdoor.py
jdgage: I don't think I follow this - why do we expect the other address?

Glenn-Clarke: The other one should also be expected, because it's the newest of its UPRN. So really both should definitely appear if the grouping is working as expected

I'm getting a bit lost reading this whole thing back, so to reset, as the test is written now:
- The sample data contains 3 EPCs. Two for one address, one for another.
- We expect this page to show two radio button entries, one for each of the two UPRNs.
- We expect neither of the two radios has the LMK matching the old EPC for the property that has two EPCs.

This should show that we're getting the newest EPC for each UPRN.

jdgage: Ok - would it be clearer what we're testing if there was only the two for the one address in the test data? Feels like we've introduced some noise to the test without needing to: ideally unit tests should test just one thing, so there's not a risk of them failing when something else has gone wrong (making it harder to debug that thing).

jdgage (approved):
Approving so as not to hold up releasing the fix while I quibble about tests

tests/test_frontdoor.py
Glenn-Clarke: I've noticed 11 Acacia never appears, at least by address. I'm running the tests quickly to see if taking it out breaks anything.

PR #403, opened 2024-11-25
PC-1429: Open EPC API Error handling

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1429)

# Description

adds some error handling to the epc interface

- 404: this was confirmed to mean no recommendations
  - don't log as this is reasonably expected
  - default to no recommendations
- 500: try up to 3 times if 500 responses are returned
  - log the error. if there are many 500 errors we would like to know
  - default to no recommendations to allow the user to continue through the flow

# Checklist

### DIFF
--- a/help_to_heat/locale/cy/LC_MESSAGES/django.po
+++ b/help_to_heat/locale/cy/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2024-11-25 11:59+0000\n"
+"POT-Creation-Date: 2024-11-25 15:19+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -671,7 +671,7 @@ msgstr "Fflat"
 msgid "Maisonette"
 msgstr "Fflat deulawr"
 
-#: help_to_heat/frontdoor/views.py:667 help_to_heat/frontdoor/views.py:737
+#: help_to_heat/frontdoor/views.py:667 help_to_heat/frontdoor/views.py:735
 msgid "I cannot find my address, I want to enter it manually"
 msgstr "Dwi'n methu gweld fy nghyfeiriad i. Hoffwn ei roi fy hunan."
 


pyproject.toml
```diff
@@ -63,7 +63,7 @@ phonenumbers = "^8.13.38"
 
 [tool.black]
 line-length = 120
-target-version = ['py38']
+target-version = ['py311']
```
samyou-softwire, line 66: needed to allow for the `match` syntax to be parsed

help_to_heat/frontdoor/interface.py
```diff
@@ -613,9 +616,40 @@ def get_address_and_epc_lmk(self, building_name_or_number, postcode):
...
+        for i in range(self.RECOMMENDATION_RETRIES_COUNT):
+            try:
+                return epc_api.get_epc_recommendations(lmk)["rows"]
+            except RequestException as requestException:
+                match requestException.response.status_code:
+                    case 500:
+                        # if on the last try
+                        if i == self.RECOMMENDATION_RETRIES_COUNT - 1:
+                            # raise this to logs, though let user continue with no recommendations
+                            logger.exception(requestException)
```
samyou-softwire, line 641: could go either way on this behaviour, though I think logging & returning blank allows the user to proceed while giving us visibility if too many errors occur

help_to_heat/frontdoor/interface.py
```diff
@@ -613,9 +616,40 @@ def get_address_and_epc_lmk(self, building_name_or_number, postcode):
...
+                            return []
+
+                        # else if not on last try, try again
+                        continue
+
+                    # 404 means no recommendations, so no need to retry
```
jdgage, line 647: ...but we want to swallow the exception as it's expected behaviour from the API. In general I think it's worth documenting why we're doing this and what we've learned about the API.

help_to_heat/frontdoor/interface.py
```diff
@@ -613,9 +616,40 @@ def get_address_and_epc_lmk(self, building_name_or_number, postcode):
...
+        epc_recommendations = self._get_epc_recommendations(epc_api, lmk)
+        return (
+            epc_details_response,
+            epc_recommendations,
+        )
+
+    def _get_epc_details(self, epc_api, lmk):
+        return epc_api.get_epc_details(lmk)["rows"][0]
+
+    RECOMMENDATION_RETRIES_COUNT = 3
```
jdgage, line 629: Would probably bump up to 5, but that's just my instinct

help_to_heat/frontdoor/interface.py
```diff
@@ -613,9 +616,40 @@ def get_address_and_epc_lmk(self, building_name_or_number, postcode):
...
+                match requestException.response.status_code:
+                    case 500:
+                        # if on the last try
+                        if i == self.RECOMMENDATION_RETRIES_COUNT - 1:
+                            # raise this to logs, though let user continue with no recommendations
+                            logger.exception(requestException)
+                            return []
+
+                        # else if not on last try, try again
+                        continue
```
jdgage, line 645: I think we should have some exponential backoff here - there's a good article [here](https://encore.dev/blog/retries) which explains why it's a good idea. It's probably not a strict necessity at this scale, and I don't think we need to worry about jitter, but it's good practice and also gives us some more resilience against short-term problems with the API.

help_to_heat/frontdoor/interface.py
```diff
@@ -613,9 +616,40 @@ def get_address_and_epc_lmk(self, building_name_or_number, postcode):
...
+            epc_recommendations,
+        )
+
+    def _get_epc_details(self, epc_api, lmk):
+        return epc_api.get_epc_details(lmk)["rows"][0]
+
+    RECOMMENDATION_RETRIES_COUNT = 3
+
+    def _get_epc_recommendations(self, epc_api, lmk):
+        for i in range(self.RECOMMENDATION_RETRIES_COUNT):
```
jdgage, line 632: Not really necessary to change, but I'd instinctively have expected this sort of thing to be a while loop with an incrementing counter. Suppose it means we don't have to create this range (technically more scalable if we were doing colossal numbers of retries).

help_to_heat/frontdoor/interface.py
jdgage: Do you mean vs throwing the error? I think this is more correct in that case: it's something we're expecting might happen, and we're handling it here rather than replying on something else further up the chain,

jdgage (changes requested):

help_to_heat/frontdoor/interface.py
samyou-softwire: have added some

help_to_heat/frontdoor/interface.py
samyou-softwire: for what it's worth range in python 3 [is lazy](https://stackoverflow.com/questions/23725258/does-range-involve-eager-or-lazy-evaluation-in-2-x-and-3-x), so it'll probably be using a counter behind the scenes. I don't think it'll ever construct the range unless we call `list()` or similar?

jdgage: Yeah I was wondering whether it ends up being equivalent once you get down to internals.

pyproject.toml
jdgage: Oh we should have changed this anyway. Would you mind have a quick look around the codebase for any other references to old Python versions? I can see there's one to 3.9 in the README, for example.

tests/test_frontdoor.py
```diff
@@ -2112,6 +2122,56 @@ def test_epc_page_shows_epc_info():
     assert page.has_one("p:contains('23 July 2010')")
 
 
+@unittest.mock.patch("help_to_heat.frontdoor.interface.EPCApi", MockRecommendationsNotFoundEPCApi)
+def test_on_recommendations_not_found_response_recommendations_are_empty_list():
+    _get_empty_recommendations_journey()
+
+
+@unittest.mock.patch("help_to_heat.frontdoor.interface.EPCApi", MockRecommendationsInternalServerErrorEPCApi)
+def test_on_recommendations_internal_server_error_response_recommendations_are_empty_list():
```
jdgage, line 2131: Would it be easy to have a test asserting specifically that we retry? e.g. mock an error the first time and then resolve the second?

samyou-softwire: don't think so; the EPC API classes are stateless (theyre reconstructed each time theyre used) & using some sort of global state could have issues with this state leaking out of the test

pyproject.toml
samyou-softwire: had a look through and I think that's the only remaining reference, so have fixed both

tests/test_frontdoor.py
jdgage: I was imagining something like
```
class MockRecommendationsTransientInternalServerErrorEPCApi(MockEPCApi):
    number_of_calls = 0
    def get_epc_recommendations(self, lmk_key):
        number_of_calls++
        if number_of_calls == 1:
            raise InternalServerErrorRequestException()

        return something normal
```

Would that work? We're retrying calls on the same instance.

samyou-softwire: ah yes it can store an epc api per request, have added one now

jdgage (approved):

PR #404, opened 2024-11-26
PC-1425: Fix for adding a space after address search first line can show no results

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1425)

# Description

adds a `.strip()` before passing off address and postcode to OS Places API

this should remove all kinds of whitespace

# Checklist

### DIFF
--- a/help_to_heat/frontdoor/views.py
+++ b/help_to_heat/frontdoor/views.py
@@ -619,8 +619,13 @@ class AddressView(PageView):
         reset_epc_details(session_id)
         session_data = interface.api.session.get_session(session_id)
         country = session_data.get(country_field)
-        building_name_or_number = data.get(address_building_name_or_number_field)
-        postcode = data.get(address_postcode_field)
+        building_name_or_number = data.get(address_building_name_or_number_field).strip()
+        postcode = data.get(address_postcode_field).strip()
+
+        # overwrite the answers with stripped ones
+        data[address_building_name_or_number_field] = building_name_or_number
+        data[address_postcode_field] = postcode
+
         try:
             if country != country_field_scotland:
                 address_and_lmk_details = interface.api.epc.get_address_and_epc_lmk(building_name_or_number, postcode)
@@ -716,9 +721,9 @@ class EpcSelectView(PageView):
 @register_page(address_select_page)
 class AddressSelectView(PageView):
     def build_extra_context(self, request, session_id, page_name, data, is_change_page):
-        data = interface.api.session.get_answer(session_id, address_page)
-        building_name_or_number = data[address_building_name_or_number_field]
-        postcode = data[address_postcode_field]
+        address_data = interface.api.session.get_answer(session_id, address_page)
+        building_name_or_number = address_data[address_building_name_or_number_field]
+        postcode = address_data[address_postcode_field]
         addresses = interface.api.address.find_addresses(building_name_or_number, postcode)
 
         uprn_options = tuple(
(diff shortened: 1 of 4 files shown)


help_to_heat/frontdoor/views.py
```diff
@@ -718,9 +718,9 @@ def save_post_data(self, data, session_id, page_name):
 @register_page(address_select_page)
 class AddressSelectView(PageView):
     def build_extra_context(self, request, session_id, page_name, data, is_change_page):
-        data = interface.api.session.get_answer(session_id, address_page)
-        building_name_or_number = data[address_building_name_or_number_field]
-        postcode = data[address_postcode_field]
+        address_data = interface.api.session.get_answer(session_id, address_page)
+        building_name_or_number = address_data[address_building_name_or_number_field].strip()
```
samyou-softwire, line 722: could possibly go more expansive on this (ie trimming the address & postcode before saving to the session), though wondered whether this would be worthwhile overall. depends on whether we want to trim the input before sending to EPC API too (though testing shows this is unnecessary)

help_to_heat/frontdoor/views.py
```diff
@@ -718,9 +718,9 @@ def save_post_data(self, data, session_id, page_name):
 @register_page(address_select_page)
 class AddressSelectView(PageView):
     def build_extra_context(self, request, session_id, page_name, data, is_change_page):
-        data = interface.api.session.get_answer(session_id, address_page)
```
samyou-softwire, line 721: renamed this as we probably shouldn't be overwriting the passed variable

Glenn-Clarke (approved):

help_to_heat/frontdoor/views.py
jdgage: Would probably vote for doing that tbh (trimming after input), just so it's not something we ever have to worry about.

samyou-softwire: have reworked to do this now

help_to_heat/frontdoor/views.py
```diff
@@ -621,9 +621,15 @@ def save_post_data(self, data, session_id, page_name):
         country = session_data.get(country_field)
         building_name_or_number = data.get(address_building_name_or_number_field)
         postcode = data.get(address_postcode_field)
```
jdgage, line 623: Life easier if you
```suggestion
        building_name_or_number = data.get(address_building_name_or_number_field).strip()
        postcode = data.get(address_postcode_field).strip()
```
up here?

jdgage (commented):
Can we cover this with tests?

help_to_heat/frontdoor/views.py
samyou-softwire: perhaps? though would lead to the line
```py
data[address_building_name_or_number_field] = building_name_or_number
```
below, which seems a bit unclear

samyou-softwire:
> Can we cover this with tests?

have added one for the main flow. can look into further testing with the OS API if we think it's worth though this'd require some more time & we're pretty much out on this ticket

help_to_heat/frontdoor/views.py
jdgage: Why's that unclear? Don't think it's any stranger than what we're doing at the moment.

Does trigger the question of whether there's an earlier point we could be doing this, before they get attached to the data object? Assuming not?

samyou-softwire: > Why's that unclear? Don't think it's any stranger than what we're doing at the moment.

just that the line `data[postcode] = postcode` may not appear to do anything useful at first. suppose can add a comment to make this clear

> Does trigger the question of whether there's an earlier point we could be doing this, before they get attached to the data object?

not really; we can specify a preprocessor for the answer schema itself but its of not much help here since this code runs before the schema is used

jdgage: Yeah I think it makes sense: we take the data, format it, and write it back.

jdgage (approved):

PR #411, opened 2024-12-11
PC-1464: Improve select your supplier question

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1464)

# Description

adds an alternative supplier page

if the user selects that they don't know they are sent here and can select any supplier

needs some extra work when displaying supplier name to ensure the correct supplier is always shown, as well as some additional work when submitting so energy suppliers see the alternative that was picked

# Checklist

# Screenshots

image
image
image
image


samyou-softwire:
noticed I've missed error texts, will push a fix

help_to_heat/frontdoor/interface.py
```diff
@@ -162,37 +169,52 @@ class SupplierConverter:
...
 
-    def _get_supplier(self):
-        return api.session.get_answer(self.session_id, "supplier")["supplier"]
+    def get_supplier(self):
+        supplier = api.session.get_answer(self.session_id, supplier_page).get(supplier_field)
+        alternative_supplier = api.session.get_answer(self.session_id, alternative_supplier_page).get(
+            alternative_supplier_field
+        )
+
+        return alternative_supplier if self._did_specify_alternative() else supplier
```
jdgage, line 178: I don't think there's real overhead to `api.session.get_answer` is there? We could be quite a bit more efficient with how many times we're calling it if so.

help_to_heat/frontdoor/interface.py
```diff
@@ -162,37 +169,52 @@ class SupplierConverter:
...
+                alternative_supplier_field
+            )
+            session_data[supplier_field] = alternative_supplier
+
         if self._is_bulb() or self._is_shell():
-            session_data["supplier"] = "Octopus Energy"
+            session_data[supplier_field] = "Octopus Energy"
         if self._is_utility_warehouse():
-            session_data["supplier"] = "E.ON Next"
+            session_data[supplier_field] = "E.ON Next"
```
jdgage, line 217: Should these be using constants?

help_to_heat/frontdoor/routing/forwards_routing.py
```diff
@@ -246,9 +252,8 @@ def _country_next_page(answers):
     return _unknown_response
 
 
-@_requires_answer(supplier_field)
-def _supplier_next_page(answers):
-    supplier = answers.get(supplier_field)
+# where to send the user post them picking a supplier
+def _post_supplier_pick_next_page(supplier):
```
jdgage, line 256: Can this be named better so the comment isn't necessary?

help_to_heat/frontdoor/consts.py
```diff
@@ -104,7 +108,7 @@
 supplier_field_shell = "Shell"
 supplier_field_utilita = "Utilita"
 supplier_field_utility_warehouse = "Utility Warehouse"
-supplier_fields = [
+supplier_fields_real = [
```
jdgage, line 111: Can this be made to read a bit more naturally? Would quibble that these aren't fields, they're values

jdgage (changes requested):

help_to_heat/frontdoor/interface.py
samyou-softwire: yep

help_to_heat/frontdoor/interface.py
samyou-softwire: don't think so too much, though it is a database call. have reworked it to just check once

help_to_heat/frontdoor/routing/forwards_routing.py
samyou-softwire: looking back I don't think the comment adds much anyway. have removed it & changed the name slightly

help_to_heat/frontdoor/consts.py
samyou-softwire: for what its worth most of these fields could be named as such, ie `supplier_field_utilitia` -> `supplier_field_value_utilita`. for now have renamed the `fields` to `field_values`

samyou-softwire:
translations for "Select an alternative energy supplier" have been added

samyou-softwire:
@jdgage if there's nothing else & pipelines pass this should be ready to go

jdgage (approved):

PR #418, opened 2024-12-30
PC-1294: Fix CSRF Fail

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1294)

# Description

use an explicit csrf origin. this should match the provided failing origin: `https://dev.check-eligibility-for-gb-insulation-scheme.service.gov.uk`

# Checklist

### DIFF
--- a/help_to_heat/settings.py
+++ b/help_to_heat/settings.py
@@ -258,7 +258,8 @@ OPEN_EPC_API_BASE_URL = env.str("OPEN_EPC_API_BASE_URL")
 
 TOTP_ISSUER = "Help to Heat Supplier Portal"
 
-CSRF_TRUSTED_ORIGINS = [BASE_URL]
+# origins don't have a trailing slash, the BASE_URL does so must be trimmed
+CSRF_TRUSTED_ORIGINS = [BASE_URL.rstrip("/")]
 
 if not DEBUG:
     SESSION_COOKIE_SECURE = True


help_to_heat/settings.py
```diff
@@ -258,7 +258,7 @@ def show_toolbar(request):
 
 TOTP_ISSUER = "Help to Heat Supplier Portal"
 
-CSRF_TRUSTED_ORIGINS = [BASE_URL]
+CSRF_TRUSTED_ORIGINS = ["https://*.check-eligibility-for-gb-insulation-scheme.service.gov.uk"]
```
jdgage, line 261: This is probably fine, but it would allow requests between environments, which isn't ideal. And it should probably come from config rather than being hard-coded.

If it was the trailing slash that was the problem, can we trim that from the base URL?

help_to_heat/settings.py
```diff
@@ -258,7 +258,7 @@ def show_toolbar(request):
 
 TOTP_ISSUER = "Help to Heat Supplier Portal"
 
-CSRF_TRUSTED_ORIGINS = [BASE_URL]
+CSRF_TRUSTED_ORIGINS = [BASE_URL.rstrip("/")]
```
jdgage, line 261: Would quite like a comment explaining why we need to trim it, if it's not obvious

jdgage (approved):
Happy as is, but a comment would be nice

PR #459, opened 2025-03-26
PC-1770: Add postcode to OS API logs

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1770)

# Description

this will help us identify any patterns in postcodes that are causing us ratelimit issues

# Checklist

# Screenshots

image

### DIFF
--- a/help_to_heat/frontdoor/os_api.py
+++ b/help_to_heat/frontdoor/os_api.py
@@ -28,16 +28,23 @@ class OSApi:
                 return self.perform_request(url)
 
             except requests.exceptions.HTTPError or requests.exceptions.RequestException as e:
+                # we log the postcode here in case of error to help diagnose if there are patterns in the kinds of
+                # postcodes that cause a ratelimit hit
+                # surrounding logs will only contain the session ID and general flow throughout the site
+                # so, it is not possible to attribute the postcode to any other details
+                # the logs themselves are retained for a week in CloudWatch
                 status_code = e.response.status_code
                 if status_code == HTTPStatus.TOO_MANY_REQUESTS:
-                    logger.error(f"The OS API usage limit has been hit for API key at index {index}.")
+                    logger.error(
+                        f"The OS API usage limit has been hit for API key at index {index} for postcode {postcode}."
+                    )
                     if index == len(self.keys) - 1:
-                        logger.error("The OS API usage limit has been hit for all API keys")
+                        logger.error(f"The OS API usage limit has been hit for all API keys for postcode {postcode}.")
                         raise ThrottledApiException
                     else:
                         continue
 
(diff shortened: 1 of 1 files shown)


help_to_heat/frontdoor/os_api.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, keys):
 
     def get_by_postcode(self, postcode, offset, max_results):
         for index, key in enumerate(self.keys):
-            url = f"""https://api.os.uk/search/places/v1/postcode?maxresults={max_results}
+            url = f"""https://api.os.uk/search/places/v1/postcode?maxrfesults={max_results}
```
Glenn-Clarke, line 22: What's happening here? Is this a typo?

Glenn-Clarke (changes requested):

help_to_heat/frontdoor/os_api.py
samyou-softwire: oops

Glenn-Clarke (approved):

jdgage (commented):
We should do some due diligence here and document it:
1. Postcodes are generally not going to be PII on their own because they aren't unique per property. However they quite quickly become PII when combined with other info. Are we logging anything else that could identify a user?
2. How long are logs stored for? I think 7 days?
3. We should be using the minimal amount of data we need: do we need the whole postcode? Could we achieve the same goal by dropping the last two or three characters of the postcode?

samyou-softwire:
> Postcodes are generally not going to be PII on their own because they aren't unique per property. However they quite quickly become PII when combined with other info. Are we logging anything else that could identify a user?

surrounding logs contain the session id and general path through the website. you could use the session ID to query the database for PII, but then again you can do this to find the postcode anyway, so this log line alone cannot be used to find other info

> How long are logs stored for? I think 7 days?

yes data retention is set to 1 week

> We should be using the minimal amount of data we need: do we need the whole postcode? Could we achieve the same goal by dropping the last two or three characters of the postcode?

ideally no. the idea behind this is to be able to discover which postcodes are causing abnormal api behaviour, so we'd ideally want the unredacted version. after testing initially it seems that in such cases the postcode isn't always written to the database, so we can't rely on that

jdgage:
Ok - I think it would be good to document some of this around the code. Don't need to worry about querying the database: if you've got access to that then you can see everyone's data anyway. Important thing is we don't have PII in the logs.

samyou-softwire:
> Ok - I think it would be good to document some of this around the code. Don't need to worry about querying the database: if you've got access to that then you can see everyone's data anyway. Important thing is we don't have PII in the logs.

have added words to this effect

jdgage (approved):

PR #478, opened 2025-04-14
PC-1762: Switch to a non-root user in our container

samyou-softwire:
# Description

adds a user to the container dockerfile

this must appear after copying the app so the user can be given ownership of it

this must appear before the `collectstatic` line else the user won't own the result files

to flag, after this is merged it may be necessary to delete your 'staticfiles' folder locally. the start script will regenerate this with the correct permissions when next run

# Checklist

### DIFF
--- a/docker/web/Dockerfile
+++ b/docker/web/Dockerfile
@@ -35,6 +35,9 @@ RUN pip install poetry && \
 
 COPY ./ /app/
 
+RUN groupadd -r appuser && useradd -r -g appuser appuser && chown -R appuser:appuser /app
+USER appuser
+
 RUN \
     DJANGO_SETTINGS_MODULE=help_to_heat.settings_base \
     DJANGO_SECRET_KEY="temp" \


samyou-softwire:
to be merged as part of [PC-1851](https://beisdigital.atlassian.net/browse/PC-1851)

docker/web/Dockerfile
```diff
@@ -35,6 +35,9 @@ RUN pip install poetry && \
 
 COPY ./ /app/
 
+RUN groupadd -r appuser && useradd -r -g appuser appuser && chown -R appuser:appuser /app
+USER appuser
+
```
jdgage, line 40: Tbh I'd do this as early as possible, which possibly is after the `chmod` commands?

samyou-softwire: we have to run at least the `chown -R appuser:appuser /app` here, otherwise there'll be no /app to edit. I could split this run command in two, create the user earlier then chown here?

jdgage: We should be able to create `/app` earlier and change its ownership at that point?

Glenn-Clarke (approved):

jdgage (approved):

PR #493, opened 2025-05-15
PC-1873: Update copyright licence to include crown copyright

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-1873)

# Description

updated years and attribute as recommended to GDS. also rename file to 'licence' as GDS recommends as well

https://gds-way.digital.cabinet-office.gov.uk/manuals/licensing.html

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 i-dot-ai
+Copyright (c) 2023 Crown Copyright (Government Digital Service)
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


XakSenSoftwire (approved):

jdgage (approved):

LICENCE
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2023 i-dot-ai
+Copyright (c) 2023-2025 Crown Copyright (Government Digital Service)
```
jdgage, line 3: Is the year range right here? I'm not a copyright lawyer, but it reads like it could be something different to e.g. just `2025 Crown Copyright`

samyou-softwire: makes sense yeah, I suppose the wording is "Where the code is continually updated with significant changes, you can show the year as a period from first to most recent update, for example 2015-2019.", though we're never really making 'significant' changes vs minor improvements, and yea it's a recommendation

jdgage (approved):

PR #525, opened 2025-08-12
PC-2010: Offboard E energy

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-2010)

# Description

as requested by DESNZ hide the ES from the list

add some structure for hidden suppliers 

# Checklist

# Screenshots

image


XakSenSoftwire (approved):
LGTM 👍 Changes & screenshot look good!

help_to_heat/frontdoor/views.py
```diff
@@ -267,6 +268,10 @@
 # to be updated when we get full list of excluded suppliers
 converted_suppliers = ["Bulb, now part of Octopus Energy", "Utility Warehouse"]
 unavailable_suppliers = []
+hidden_suppliers = [supplier_field_e]
+visible_supplier_options = [
+    supplier_option for supplier_option in schemas.supplier_options if supplier_option["value"] not in hidden_suppliers
```
LeoCie, line 273: Do you think it may be worth doing something like deprecating schemas.supplier_options?
I guess we still need it in some other logic in the app / for old submissions?
I'm worried that it may get used elsewhere without knowledge that visible... is the user-facing option. Maybe just a comment? Happy to be overruled if you think it's unlikely!

samyou-softwire: I think the intent is that `schemas.py` contains structures to do with validation. we can't really deprecate `supplier_options` as we use it later on in `schemas.py` to validate that the user has answered one of the suppliers

`views.py` contains structures to do with the user facing side so I put the code here, though agree this isn't the most clear. I'll comment the schema

PR #537, opened 2025-09-18
PC-2049: Redirect main site users to start page

samyou-softwire:
[Link to Jira ticket](https://beisdigital.atlassian.net/browse/PC-2049)

# Description

adds a new database model to store live settings that we can update without restarting the app

adds overrides to the following pages to check the public site open setting and redirect to the start page if its on:
- `start`
- `<uuid:session_id>/<str:page_name>/`
- `<uuid:session_id>/<str:page_name>/change/`

other pages in the app are static and so should be fine to let users still view these (such as privacy policy)

adds a CLI command to turn the public portal block on and off

running of these tasks is documented here: https://softwiretech.atlassian.net/wiki/spaces/Support/pages/21622325447/DESNZ+Support+GBIS+Closure+Runbook

# Checklist


help_to_heat/frontdoor/live_settings.py
```diff
@@ -0,0 +1,30 @@
...
+def set_public_portal_open(open):
+    live_settings = _get_or_create_live_settings()
+    live_settings.public_portal_open = open
+    _save_live_settings(live_settings)
+
+
+def _get_or_create_live_settings():
+    if not LiveSettings.objects.exists():
+        live_settings = LiveSettings()
+        _save_live_settings(live_settings)
```
samyou-softwire, line 22: this is potentially dangerous if two users try to generate live settings at once, could lead to two rows in this table. have added to ticket release notes to check the database manually & ensure this is not the case

once it exists in the db we never generate a new LiveSettings object so should be safe going into the future

help_to_heat/frontdoor/models.py
```diff
@@ -48,3 +48,7 @@ class Feedback(utils.UUIDPrimaryKeyBase, utils.TimeStampedModel):
 
     def __str__(self):
         return f"<feedback id={self.id} page_name={self.page_name}>"
+
+
+class LiveSettings(utils.TimeStampedModel):
```
samyou-softwire, line 53: happy to hear on different implementations of this model (a table with a single row with a col for each setting, a table with a row for each settings or something else)

XakSenSoftwire: Summarised adv & disadv (ChatGPT)

Wide table - one row, many columns (current):

- Strong typing: Each flag is a real column with a boolean type → database enforces valid values.
- Clear code: LiveSettings.objects.first().public_portal_open is clear & self-describing and IDEs can autocomplete the field.
- One query for everything: Fetching one row gives you all flags.
- Migration required for each new flag: schema must change every time you add one.
- Not dynamic: Harder to add flags on the fly (you must deploy code + run migration).

Tall table - one row per flag:

- No schema changes needed: Adding a new flag = just insert a new row.
- Dynamic: You could even add/remove flags at runtime from an admin UI or CLI.
- Scales better: If you have dozens or hundreds of flags, this avoids table bloat.
- Weaker typing: name is just a string → no DB-level enforcement that a flag exists.
- Slightly slower: Fetching all flags requires querying multiple rows and mapping them into a dictionary.
- Easier to mis-spell: Access becomes FeatureFlag.objects.get(name="public_portal_open"), and a typo will silently fail.
- More moving parts in code: You usually write helper functions to fetch flags safely.

---

Given that we don't have the plan to have many feature flags & don't have the requirement to create the flags dynamically I think the wide table is the best option 👍

XakSenSoftwire: Instead of doing this with a db, we could look at using an AWS service to manage the feature flags such as AWS AppConfig. These should work similarly (i.e. don't need to restart the app when toggling the flag) to the db implementation but also have the additional upside of rolling out the change (i.e. maintenance mode) gradually to users. I don't think this is something we need so we should proceed with the db implementation

help_to_heat/frontdoor/views.py
```diff
@@ -326,6 +330,9 @@ def not_found_page_view(request, exception):
 
 
 def page_view(request, session_id, page_name):
```
XakSenSoftwire, line 332: This current implementation relies on the user navigating to a new page to see that the service has just gone into maintenance mode. I think this could be problematic as we could:

- user is on page to input some information to the DB
- user walks away from their computer
- we put the site in maintenance mode
- user comes back to finish form
- submits data
- unintentional behaviour by inputting data into db whilst in maintenance mode

Not sure if this is an issue but a thought I had

XakSenSoftwire: Bogdan mentioned to do something like scheduling maintenance time with the CLI command. We could add a frontend notification saying "The site is going to be in maintenance mode in X hours - your answers will not be saved after this time". Then we just  make sure maintenance mode makes sure no calls to the database are being made.

XakSenSoftwire (commented):
Nice work - left comments :)

help_to_heat/portal/management/commands/set_public_portal_open.py
```diff
@@ -0,0 +1,73 @@
...
+            case "closed":
+                arg_open = False
+
+        # query the current public portal state
+        live_setting_open = public_portal_is_open()
+
+        print("!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!")
+        print("This is a dangerous command!")
+        print("Running this command toggles whether the public portal of the site is closed or open.")
+        print("If the public portal is closed, users will not be able to submit referrals.")
```
XakSenSoftwire, line 31: Have we specifically enforced this somewhere? That db calls cannot be made when this flag is on.

XakSenSoftwire (commented):
Left another as I forgot to review this - oops!

help_to_heat/frontdoor/views.py
samyou-softwire: > This current implementation relies on the user navigating to a new page to see that the service has just gone into maintenance mode. I think this could be problematic as we could:

my understanding is that the page_view overrides will handles both GET and POST requests so any POST requests while the site is closed will hit the redirect and not go to the site

I gave it a test locally and this does seem to be true, though interestingly content secure policies prohibit the user from submitting the form entirely while the site is closed. pressing 'submit' just does nothing (no redirect at all). leaves us with some options:

- leave as is (user will eventually press back or refresh and see the gov.uk page)
- add an exception to the content secure policies to allow POSTing to GOV.UK (potential security risk but we can be confident of the addresses all our forms POST to, the links are not dynamic)
- add some new page that 'handles' the post and then redirects (cleanest but would take the most time)

I've added a commit to update the CSP list, though happy to look into other options

> Bogdan mentioned to do something like scheduling maintenance time with the CLI command.

I think a notice like that would be good. as we're certain of the closure date probably doesn't need to be more than a static sign. will check with DESNZ

help_to_heat/portal/management/commands/set_public_portal_open.py
samyou-softwire: as mentioned above all POST calls to page views will be redirected & so don't save new data

XakSenSoftwire: makes sense - thanks for clarifying

help_to_heat/frontdoor/live_settings.py
davidm-m: I'm not going to ask to change this because I agree, once it's set up the risk is gone, but there should be a way to acquire a lock on the table to prevent two people from trying to do this at once.

davidm-m (approved):
All looks good - are we going to add a banner warning users of the upcoming closure as part of this or as a separate ticket? I don't think there's any risk in this being merged before that's done anyway.

On the CSP change, adding gov.uk isn't a risk IMO and I agree it's a nicer user experience.

samyou-softwire:
> All looks good - are we going to add a banner warning users of the upcoming closure as part of this or as a separate ticket? I don't think there's any risk in this being merged before that's done anyway.
> 
> On the CSP change, adding gov.uk isn't a risk IMO and I agree it's a nicer user experience.

sounded like a separate ticket last we asked (as it's blocked on GDS deciding how they will change the start page)

